PR #86, opened 2022-04-26
Issue 84

### LINKED ISSUE #84: Restore HTTPS download functionality to data subscriber script

maseca:
Closes #84


cluster_provisioning/modules/common/variables.tf
```diff
@@ -130,12 +130,12 @@ variable "amis" {
   type = map(string)
   default = {
     # HySDS v4.0.1-beta.8-oraclelinux
-    mozart    = "ami-02fcd254c71ff0fa0"  # opera dev mozart - ol8
-    metrics   = "ami-0a54a14946e0bb52f"  # opera dev metrics - ol8
-    grq       = "ami-0a11c7d42e24fe7d5"  # opera dev grq - ol8
-    factotum  = "ami-0ce5e6a66b7732993"  # opera dev factotum - ol8
-    ci        = "ami-0caed57c920d65ea8"  # OL8 All-project verdi v4.11
-    autoscale = "ami-0caed57c920d65ea8"  # OL8 All-project verdi v4.11
+    mozart    = "ami-031dd1ada8418ff92" # mozart v4.16-1 opera dev
```
riverma, line 133: The public (or anyone else using the code) won't have access to these AMIs. As a best practice, its best to leave specific environment configuration like this out of revision control. 

Couple suggestions:
- Let's keep configuration like this out of our public GitHub, and version control internally with our deployment scripts
- Just mention in the README that things like this need to be customized

maseca: Hi Rishi,

You should start a thread elsewhere about this so that it can be dealt with in a future ticket. I had to update the existing values for my development deployment to work. This PR is not intended to focus on AMIs. I would recommend tagging @hhlee445 as well.

riverma: Sounds good @maseca - will do thanks 👍

PR #90, opened 2022-04-27
DSWx-HLS Beta PGE Integration

collinss-jpl:
This pull request makes the required changes to support integration of the DSWx-HLS PGE built with the "beta" delivery of the corresponding SAS. High level changes include:

- Updates to terraform scripts to point at the "beta preview" build for the DSWx-HLS PGE for the main, dev, dev-e2e, dev-int and int profiles.
- Updates to SDS config files to support runconfig format changes and creation of multiple output tif files by the PGE
- Refactoring of the `product2dataset` and `extractor` modules to strip unneeded NISAR functionality, as well as support bundling of multiple tif output bands into a single SDS Dataset


hhlee445:
@collinss-jpl Can we also have a simulated L3_DSWx_HLS PGE to produce multiple tif files?

PR #113, opened 2022-05-25
opera-sds-int-issues-13: First iteration of the crontab file generato…

philipjyoon:
…r. Haven't had time to validate yet.

### DIFF
--- a/load_testing/scripts/gen_cron_daac_data_subscriber.py
+++ b/load_testing/scripts/gen_cron_daac_data_subscriber.py
@@ -53,7 +53,7 @@ date_format_str = "%Y-%m-%dT%H:%M:%SZ"
 increment_mins = args.job_period * args.multiplier
 
 start_dt = datetime.strptime(args.start_date, date_format_str)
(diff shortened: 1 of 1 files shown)


philipjyoon:
Looks like the last of the 5 values for cron schedule is day of week (which I should just use * for) instead of the year. I got it be the year from bad source it seems. Will make that fix.

philipjyoon:
Time format is also incorrect. And I think the start time is off. These are quick fixes and should be able to fix them all today.

philipjyoon:
Updated to match the latest daac_data script and also simplified to take pass-through arguments. Example usage:

./gen_cron_daac_data_subscriber.py -m query -s 2022-01-17T10:00:00Z -jp 5 -jc 2 -pass "-c HLSS30 -p LPCLOUD -i opera-int-isl-fwd --release-version=1.0.0-er.4.0 --job-queue=opera-job_worker-data_subscriber_download --chunk-size=80 --smoke-run --dry-run"

hhlee445 (approved):

Issue #108, opened 2022-05-19
[New Feature]: Configurable CMR endpoints


hhlee445:
### Checked for duplicates

Yes - I've already checked

### Alternatives considered

Yes - and alternatives don't suffice

### Describe the feature request

Current version of daac_data_subscriber.py is using a hard coded EARTHDATA login URL (**https://urs.earthdata.nasa.gov**) to query CMR. 

We should be able to pass different EARTHDATA URL for testing. 

Possible earth data login URLs:
1. OPS: https://urs.earthdata.nasa.gov/
2. UAT: https://uat.urs.earthdata.nasa.gov

We may consider to use the file 'daac_environments.yaml'.

PR #158, opened 2022-07-19
Issue 108

hhlee445:


maseca:
Closes #108

maseca:
Note: OPS and UAT endpoint products are incompatible and any attempt to download a product from the wrong endpoint will result in an error

hhlee445 (commented):
It queries successfully and ingests into hls_catalog and hls_spatial_catalog, however, it doesn't submit download job. Don't we need to pass --endpoint to trigger download job (submit_download_job)?

maseca:
Tested on Mozart with CLI and in browser on demand

Issue #202, opened 2022-09-01
[New Feature]: Enable Simulation PGE toggle from terraform


philipjyoon:
### Checked for duplicates

Yes - I've already checked

### Alternatives considered

Yes - and alternatives don't suffice

### Describe the feature request

Currently the terraform always deploys w Simulation PGE. But the vast majority of the cases of cluster deployments require non-sim, aka real PGEs. Switching the cluster over to real-PGE is a separate and manual process that must be done after the cluster is deployments. Between the deployment often being complex and time-intensive this pge-sim-toggle step is sometimes overlooked. the cost of this manual error is high: the cluster must be stopped, all jobs cleaned up, and then restart the task-at-hand - it would easily set back several hours of the test/run taking place.

I would like to introduce a new variable called "pge_sim_mode", true by default, that will update the Mozart settings.yaml file at deployment time if it's set to false. Implementing this way it will be affect anyone else's workflow - it will just make the CM life easier and therefore the team a bit more efficient.

philipjyoon:
This issue has now been completed.

PR #225, opened 2022-09-30
Completing Issue 202 by adding pge_sim_mode to int and pst venues

philipjyoon:
This completes issue 202


riverma:
Just a tip, try always linking to the respective issue by including a `#` character before the number. i.e. #202

hhlee445 (approved):

PR #282, opened 2022-11-03
Issue 278: SLC consolidated pipeline. orbit file download

chrisjrd:
perform download+extract+ingest operations in same job. This includes bundling an orbit file with the dataset

### DIFF
--- a/cluster_provisioning/modules/common/main.tf
+++ b/cluster_provisioning/modules/common/main.tf
@@ -1598,6 +1598,24 @@ data "template_file" "launch_template_user_data" {
                     "timestamp_format": "%Y-%m-%d %H:%M:%S,%f"
                   },
                   {
+                    "file_path": "/data/work/jobs/**/run_slcs1a_query.log",
+                    "log_group_name": "/opera/sds/${var.project}-${var.venue}-${local.counter}/run_slcs1a_query.log",
+                    "timezone": "Local",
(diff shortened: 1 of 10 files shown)


cluster_provisioning/modules/common/main.tf
```diff
@@ -1632,6 +1650,18 @@ data "template_file" "launch_template_user_data" {
...
                     "timestamp_format": "%Y-%m-%d %H:%M:%S,%f"
                   },
+                  {
+                    "file_path": "/home/ops/verdi/log/opera-job_worker-slc_data_query.log",
+                    "log_group_name": "/opera/sds/${var.project}-${var.venue}-${local.counter}/opera-job_worker-slc_data_query.log",
+                    "timezone": "Local",
+                    "timestamp_format": "%Y-%m-%d %H:%M:%S,%f"
+                  },
+                  {
+                    "file_path": "/home/ops/verdi/log/opera-job_worker-slc_data_download.log",
```
riverma, line 1660: Hey @chrisjrd - this assumes we're always installing / setting up on a directory called `ops` within the folder `home`. Is there a better way we can safely write file paths like this such that if the username changes, this won't break?

riverma (commented):
I see you added the orbit collection with the single, simplified pipeline model. That's great. Are there any download / ingest steps within the pipeline *not* bundled together that are currently stand-alone? Other than querying for data sets from CMR?

Issue #308, opened 2022-11-23
Terraform deployment not working on ARM64


chrisjrd:
System deployment on arm64 platforms is broken. Terraform providers are not available for arm64 using the versions specified.

Address by making the necessary changes to enable successful system deployment. This includes using updated Terraform providers.

riverma:
@chrisjrd - are we planning to use any ARM64 processors for R1,R2?

PR #309, opened 2022-11-23
feat(terraform): fix deployment on ARM64

chrisjrd:
- removed: use of `template_file`/`template` terraform provider
- misc: updated pytest version referenced in setup.py

Refs #308


hhlee445:
This fix works with terraform 1.0.11 as well.

hhlee445:
This fix caused E2E jenkins job to fail. need to test more.

PR #352, opened 2023-01-12
Issue 350: geospatial filter for SLC + CSLC

chrisjrd:
also fixes typos in the SLC S1A and SLC S1B job specs. Typo was a misleading native-id example value. This was using an example valid only for HLS data rather than for SLC.

Refactored ElasticSearch connection management. Only a single, shared connection reference is used in subscriber. Previously, multiple connection objects were created.

fixed typo in integration tests confttest.py. Typo was for index name. Wrong index name results in improper cleanup before test execution, since impacted indexes are never dropped.


hhlee445:
Getting an error as 

```
Traceback (most recent call last):
  File "/export/home/hysdsops/mozart/ops/opera-pcm/data_subscriber/daac_data_subscriber.py", line 39, in <module>
    from geo.geo_util import does_bbox_intersect_north_america
  File "/export/home/hysdsops/mozart/ops/opera-pcm/geo/geo_util.py", line 7, in <module>
    from osgeo import ogr
ModuleNotFoundError: No module named 'osgeo'
```

hhlee445 (approved):

PR #387, opened 2023-02-10
Issue 370 more terraform refactor

philipjyoon:


hhlee445 (changes requested):
PST, INT, and OPS should not use dev-e2e/variables.tf since many variables are different from DEV environment. I recommend to create a common variables.tf for PST, INT, and OPS.
1. RS bucket retention policy is different between DEV and PST. 
2. ASG queue sizes are different between DEV and PST/INT/OPS.
3. AMIs are different from OPS and DEV.

philipjyoon:
My understanding is that variables.tf defines what _can_ change between venues; and override.tf defines what those different values are. It's almost like variables.tf declares variables in a function and overrride defines the values for the variables. 

Variables.tf doesn't have to define the default values but many do probably out of convenience. Agree that different venues can and do have different _values_ for the variables but they all _define_ and _use_ the same variable. Is this correct? In the end variables.tf just defines what the variables are being used in main.tf. We would still have one or more override.tf that's common between certain venues or specific to a venue to account for all the value differences. 

So the idea is to standardize a single variables.tf file because all venues needs the same variable file. Issues I run into is that a new variable is introduced in a build that may or may not functionally affect it and the variable is created in just some of the venues and not all. But this new variable is used in main.tf somewhere. So this makes deployment an exercise of forensics in a way. If we at least use a standardized variables.tf we would get an error at terraform plan stage instead of getting error deep into the deployment. The former will error out in seconds and can be fixed right away; the latter will error out in 40-50 mins and it will take 2+ hr cycle to fix - after figuring out what the new variable that needs to be added.

hhlee445 (commented):
@philipjyoon We should keep ops/variables.tf and ops/override.tf as separate files since it will be different from dev-e2e/variables.tf & override.tf

philipjyoon:
@hhlee445 ops/override.tf is a different file from dev-e2e/overrride.tf. It's its own file.

As for ops/variables.tf vs dev-e2e/variables.tf, the values will be different but would they also have different variables being declared? The latter seems unlikely since they're using the same modules/common/main.tf

hhlee445 (approved):

PR #400, opened 2023-02-28
Issue 373: fix SLC download job running out of disk space

chrisjrd:
fix download job running out of disk space. issue caused by orbit number not being unique enough for a batch of data that can fit onto a worker's disk space.


maseca (approved):
Deployed and queried the past 10 hours of S1A, resulting in ~2.3k download jobs which ran smoothly. Only failures were CNM jobs which are beyond the scope of this issue.

maseca:
Got the disk space error after running for about an hour. Reverted the PR

PR #403, opened 2023-03-01
Read .env file in pcm_batch.py

### LINKED ISSUE #394: pcm_batch.py and audit.py should use GRQ IP already known on the Mozart box
### Checked for duplicates

Yes - I've already checked

### Alternatives considered

Yes - and alternatives don't suffice

### Related problems

_No response_

### Describe the feature request

Right now usage of pcm_batch.py requires the user to set the IP address of GRQ as env var. But this information is already available in ~/.sds/config so we should just use that.

audit.py requires similar info so we should apply the similar pattern there.

maseca:
Closes #394

### DIFF
--- a/tools/pcm_batch.py
+++ b/tools/pcm_batch.py
@@ -5,23 +5,21 @@ from __future__ import print_function
 import argparse
 import json
 import logging
-import os
 import sys
 import time
 from datetime import datetime, timedelta
+from pathlib import Path
 from types import SimpleNamespace
 
-from dotenv import dotenv_values
 from hysds_commons.elasticsearch_utils import ElasticsearchUtility
+
+from util.conf_util import SettingsConf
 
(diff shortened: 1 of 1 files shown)


philipjyoon:
Max, the main goal of this ticket was to resuse the GRQ_IP information that already exists on the Mozart machine  ~/.sds/config so that the user does not need to look for it every time and set it as an ENV var or in the .env file.

Also we want this same code logic implemented in audit.py

maseca:
> Phil Yoon
> [11:20 AM](https://jpl.slack.com/archives/D03E1E2C3ML/p1677784812360909)
> Looks great. I'm not on my machine and don't remember my git creds off of my head so can't formally approve
> [11:20](https://jpl.slack.com/archives/D03E1E2C3ML/p1677784821462789)
> but it looks good so consider this as the approval

PR #407, opened 2023-03-06
Disable Ancillary Overlap Check for DSWx-HLS

collinss-jpl:
This PR sets the `check_ancillary_inputs_coverage` flag to False in the template RunConfig for DSWx-HLS. This check is disabled while a bug is corrected in the SAS to fix the anti-meridian case with the global Landcover file.

The underlying SAS issue is being addressed here: https://github.com/nasa/PROTEUS/pull/43


chrisjrd (approved):
looks good. if you can, put a ticket number for the SAS issue or an expected date of resolution in in the comment. It makes for a more complete comment.
But it's up to you.
As is, it is good to merge.

PR #444, opened 2023-03-27
Add margin before antimeridian crossing test

gshiroma:
This PR fixes `stage_dem.py` and `stage_worldcover.py` for the case in which the MGRS tile does not cross the antimeridian, but the added margin should result in antimeridian crossing. The fact that the MGRS tile does not cross the dateline prevents the current code from detecting the antimeridian crossing and subsequently dividing the tile polygon into two pieces.

The solution is to add the margin before the antimeridian crossing test. This PR adds the margin (in km) to the tile coordinates in meters, which may be preferable, but keeps them conversion using the meters-to-degrees conversion at the Equator (i.e., with longitude margin matching the latitude margin) as the default. This option is selected by the flag `flag_use_m_to_deg_conversion_at_equator` in the call to the function `polygon_from_mgrs_tile()`.

### DIFF
--- a/tools/stage_worldcover.py
+++ b/tools/stage_worldcover.py
@@ -173,8 +173,6 @@ def download_worldcover(polys, worldcover_bucket, worldcover_ver,
         Path to the where the output Worldcover file is to be staged.
 
     """
-    # convert margin to degree (approx formula)
-    margin = margin / 40000 * 360
 
     # Download Worldcover map for each polygon/epsg
     file_prefix = os.path.splitext(outfile)[0]
--- a/tools/stage_worldcover.py
+++ b/tools/stage_worldcover.py
@@ -259,7 +259,7 @@ def main(opts):
     if not opts.worldcover_year:
         opts.worldcover_year = WORLDCOVER_YEAR
 
-    # Determine polygon based on MGRS info or bbox
+    # Determine polygon based on MGRS grid reference with a margin, or bbox
     poly = determine_polygon(opts.tile_code, opts.bbox, opts.margin)
 
     # Check dateline crossing. Returns list of polygons
--- a/tools/stage_worldcover.py
+++ b/tools/stage_worldcover.py
@@ -67,7 +67,7 @@ def get_parser():
     return parser
 
 
-def determine_polygon(tile_code, bbox=None):
+def determine_polygon(tile_code, bbox=None, margin_in_km=50):
     """
     Determine bounding polygon using MGRS tile code or user-defined bounding box.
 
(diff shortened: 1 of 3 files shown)


tools/stage_worldcover.py
```diff
@@ -149,7 +153,7 @@ def translate_worldcover(vrt_filename, output_path, x_min, x_max, y_min, y_max):
 
 
 def download_worldcover(polys, worldcover_bucket, worldcover_ver,
-                        worldcover_year, margin, outfile):
+                        worldcover_year, outfile):
```
collinss-jpl, line 156: [Line 175 ](https://github.com/nasa/opera-sds-pcm/pull/444/files#diff-6c25f58608c1d9cb3ab55fedaef6e32fbc9a1d8ce850ff6def523ec10af14f0cL175) needs to be removed as well

gshiroma: Good catch. Thank you, @collinss-jpl !

util/geo_util.py
```diff
@@ -18,6 +22,17 @@ def polygon_from_mgrs_tile(mgrs_tile_code, margin_in_km):
     -----------
     mgrs_tile_code : str
         MGRS tile code corresponding to the polygon to derive.
+    margin_in_km : float
+        Margin in kilometers to be added to MGRS bounding box
+    flag_use_m_to_deg_conversion_at_equator : bool
```
collinss-jpl, line 27: Is there a rule of thumb when we (PCM) could use to determine when this flag should be set to True or False? Can we trigger off the particular MGRS tile code if its one near the Equator?

gshiroma: For DSWx-HLS, we can always have `flag_use_m_to_deg_conversion_at_equator` set to `True`.

I left this flag to avoid changes with the current system/processing. In theory, the right thing to do would be to have `flag_use_m_to_deg_conversion_at_equator` set to `False`. But since we want to avoid last-minute changes, I thought it would be safer to keep the previous margin sizes (i.e., with the margin in longitude computed at the equator, even if the products are not near the equator).

collinss-jpl (approved):
I've confirmed this branch now allows the T60UXB and T60WWD tiles to complete successfully in PCM. @hhlee445 this branch should be good to merge. @gshiroma thanks again for your assistance, I will be sending you an LFT with the results for the two tiles.

gshiroma:
> I've confirmed this branch now allows the T60UXB and T60WWD tiles to complete successfully in PCM. @hhlee445 this branch should be good to merge. @gshiroma thanks again for your assistance, I will be sending you an LFT with the results for the two tiles.

Thank you so much, @collinss-jpl , for all your support and for reviewing this PR. I received your email. I'm going to verify these two products now.

hhlee445:
@collinss-jpl @gshiroma do we need to include this fix on Release 1.0.0?

hhlee445 (approved):

PR #470, opened 2023-04-19
LP Token Patch

maseca:
Addresses Cumulus bug that breaks Basic Auth for s3 credentials retrieval

### DIFF
--- a/data_subscriber/daac_data_subscriber.py
+++ b/data_subscriber/daac_data_subscriber.py
@@ -1180,8 +1180,8 @@ def download_product_using_https(url, session: requests.Session, token, target_d
         return product_download_path.resolve()
 
 
-def download_product_using_s3(url, session: requests.Session, target_dirpath: Path, args) -> Path:
-    aws_creds = _get_aws_creds(session)
+def download_product_using_s3(url, token, target_dirpath: Path, args) -> Path:
+    aws_creds = _get_aws_creds(token)
     logging.debug(f"{_get_aws_creds.cache_info()=}")
 
     s3 = boto3.Session(aws_access_key_id=aws_creds['accessKeyId'],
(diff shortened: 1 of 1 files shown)


riverma:
@maseca - great work and timing here! Any changes specified to the `.netrc` file to leverage the token? Can you give a template example of how to specific the token (without placing any token here) in the `.netrc`?

maseca:
So the token handling is done via the URS token API. The `.netrc` file is only used to provide the basic auth required to interact with the token API. Once the token is retrieved by the script via API it is stored in working memory until the script terminates. If no existing valid token is present based on the results of the initial API call, a second call will be performed in order to generate a valid token.

data_subscriber/daac_data_subscriber.py
```diff
@@ -1252,10 +1252,11 @@ def _to_s3_url(dl_dict: dict[str, Any]) -> str:
 
 
 @ttl_cache(ttl=3300)  # 3300s == 55m. Refresh credentials before expiry. Note: validity period is 60 minutes
-def _get_aws_creds(session):
+def _get_aws_creds(token):
     logging.info("entry")
 
-    with session.get("https://data.lpdaac.earthdatacloud.nasa.gov/s3credentials") as r:
+    with requests.get("https://data.lpdaac.earthdatacloud.nasa.gov/s3credentials",
+                      headers={'Authorization': f'Bearer {token}'}) as r:
```
riverma, line 1259: Nicely done!

riverma (commented):
Great work @maseca - appreciate your rapid patch here. 

Questions:
- Will an operator need to specify the token explicitly somewhere? 
- How long would a token last, and is automatic renewal taken care of?

data_subscriber/daac_data_subscriber.py
LucaCinquini: I agree - great surgical changes :)

maseca:
Operators will still have access to the URS GUI for anything token-related, but they shouldn't have to manually do anything. The PCM will generate new tokens automatically via the API when they are about to expire.

hhlee445 (approved):

PR #486, opened 2023-05-04
Issue 472: improve audit tools with direct CMR queries

chrisjrd:


hhlee445:
@chrisjrd Should we rename the script to 'dswx_hls_audit.py'? then, we can create a new audit script for R2

chrisjrd:
> 

If we'll end up with a suite of audit scripts, it might be good to rename them and put them in their own module (directory). But I don't think we need to do that until we start working on the R2 audit script, when it will be clear what can be reused to share between different audits.

hhlee445 (approved):

Issue #500, opened 2023-05-16
[Bug]: hls_download jobs halted with out of memory error when using native-id query 


hhlee445:
### Checked for duplicates

Yes - I've already checked

### Describe the bug

When the OPS team submitted the missing granule jobs using the native-id parameter, the hls_download jobs were halted.

Reports from Phil and Chris,

> The issue is that every time the --native-id / --batch-ids parameters is used the OPERA download code will first query the entire GRQ ES for all download jobs from start-date and end-date. This is a design flaw in that code - there is no reason for it to query for all date range when it already has a unique value of the native-id. It should just simply query ES using that unique ID.
> Long-term solution is to fix this design flaw in a future R1 release.
> Short-term fix is to:
> 1. Purge/Revoke all these native-id jobs. These will never complete and these are getting way of normal forward processing.
> 2. We have to specify the start-date for the native-id query. Fortunately these are embedded in the native-id itself.
> 3. So what I can do is modify the script you're using to generate daac_data_sub... commands from a list of native-ids to 
> a. parse out the start-date from the native-id
> b. and then generate the start-date parameter out of that
> 
>The end result is that your process stays the same but the command you use will now have the appropriate start-date and download jobs should work using native-id.

### What did you expect?

We expected to download one granule per one download queue and continue to trigger HLS SCIFLO job.

### Environment

```bash
- Version of this software: PCM 1.0.2
-
```

philipjyoon:
Fixed as described.

To validate: first query and download a good amount of data. Then submit a single native-id query. Note that it no longer logs out paginated ES query output which would have been indicative of the large unbounded date range query. Finally note that the native-id granule was downloaded and the corresponding SCIFLO job was queued.

1. python mozart/ops/opera-pcm/data_subscriber/daac_data_subscriber.py query -s 2023-03-04T00:00:00Z -e 2023-03-04T01:00:00Z -c HLSS30 -p LPCLOUD --use-temporal --release-version=issue_500_native_id_download --job-queue=opera-job_worker-hls_data_download —chunk-size=1

2. python mozart/ops/opera-pcm/data_subscriber/daac_data_subscriber.py query --native-id HLS.S30.T55VDF.2023063T014651.v2.0 -p LPCLOUD --release-version=issue_500_native_id_download --job-queue=opera-job_worker-hls_data_download —chunk-size=1

philipjyoon:
This was fixed for R1. Now fixing for R2. To test for R2:

[...]

PR #503, opened 2023-05-19
Issue #500. When downloading batch id length is 1, do a direct ES que…

philipjyoon:
…ry for that granule id to save resources


hhlee445:
https://github.com/nasa/opera-sds-pcm/pull/509

PR #509, opened 2023-05-31
Issue #500. When downloading batch id length is 1, do a direct ES que…

hhlee445:
…ry for that granule id to save resources


philipjyoon:
Will be implementing a different, more complete, solution on a different branch. Ignore this pull request and branch for now.

PR #510, opened 2023-05-31
Improve audit tools with direct CMR queries for R1

hhlee445:


riverma:
@hhlee445 - please coordinate with @LalaP on the OPS use case for this if you can. At a minimum, we should have OPS lead review the proposed improvements to provide feedback on usage.

hhlee445:
this fix was delivered into release 2, no need to add into 1.0.3 branch

PR #525, opened 2023-06-16
Issue 478 529: SLC ionosphere file download and CSLC submission

chrisjrd:


data_subscriber/ionosphere_download.py
```diff
@@ -0,0 +1,340 @@
...
+                job_submission_results = await submit_cslc_jobs([slc_dataset], args)
+
+                if job_submission_results["success"]:
+                    results["success"].extend(job_submission_results["success"])
+                elif job_submission_results["fail"]:
+                    results["fail"].extend(job_submission_results["fail"])
+                else:
+                    pass
+
+                if results["fail"]:
```
collinss-jpl, line 90: Not sure `results["fail"]` is what you want to always be checking here. For example, if two jobs are to be submitted and the first fails (and is added to `results["fail"]`), the second job would also trigger this condition even if it was successful, since `results` is not cleared between iterations of the loop.

Maybe this should be checking `job_submission_results["fail"]` instead? In that case lines 90-93 could be merged into the conditional branch at line 85?

data_subscriber/ionosphere_download.py
```diff
@@ -0,0 +1,340 @@
...
+                    logging.info(f"Job submission failure result for {product_id=}. Collecting exception result. Skipping to next SLC dataset.")
+                    exceptions.extend(results["fail"])
```
collinss-jpl, line 92: Again, I think you want to call `extensions.extend(job_submission_results["fail"])` instead of using `results['fail']`, since `results` is cumulative, and thus could results in redundant entries in `exceptions` for every failed job.

data_subscriber/ionosphere_download.py
```diff
@@ -0,0 +1,340 @@
...
+                ionosphere_url = get_ionosphere_correction_file_url(dataset_dir, product_id)
+                logger.info(f"{output_ionosphere_filepath=}")
+                logger.info(f"{ionosphere_url=}")
+
+                slc_dataset_s3_url: str = next(iter(filter(lambda url: url.startswith("s3"), slc_dataset["_source"]["browse_urls"])))
+                logger.info(f"{slc_dataset_s3_url=}")
+
+                s3_bucket, s3_key = try_s3_upload_file(slc_dataset_s3_url, output_ionosphere_filepath)
+
+                job_submission_results = await submit_cslc_jobs([slc_dataset], args)
```
collinss-jpl, line 81: Since `submit_cslc_jobs` is set up to work on multiple job requests at a time, lines 81-93 could be moved outside of this loop and only called once using the entire list of `slc_dataset` that we got ionosphere files for.

collinss-jpl (changes requested):
Overall approach looks solid, but I had some concerns regarding the core loop over the SLC datasets in ionosphere_download.py. See my comments below.

cluster_provisioning/modules/common/variables.tf
```diff
@@ -369,6 +373,15 @@ variable "queues" {
       "total_jobs_metric" = true
       "use_private_vpc" = false
     }
+    "opera-job_worker-slc_data_download_ionosphere" = {
+      "instance_type" = ["c5n.2xlarge", "m5dn.2xlarge"]
```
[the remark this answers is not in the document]
chrisjrd: No thought. I jut copy-pasted from the top. We previously had issues with another job using medium size (probably due to RAM). We can use something like m6a.large or m6i.large to get the work done.

data_subscriber/download.py
```diff
@@ -204,34 +204,14 @@ def download_from_asf(
 
         logging.info("Added orbit file to dataset")
 
-        if additional_metadata.get("intersects_north_america", False):
-            logging.info("Downloading associated Ionosphere Correction file")
+        if additional_metadata.get("intersects_north_america", False) \
+                and (additional_metadata['processing_mode'] == "historical" or additional_metadata['processing_mode'] == "reprocessing"):
```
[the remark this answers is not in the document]
chrisjrd: I'm not sure where this came from, but I'll make the change.

data_subscriber/ionosphere_download.py
```diff
@@ -0,0 +1,340 @@
...
+            if not slc_dataset["_source"]["metadata"].get("intersects_north_america"):
+                logging.info("dataset doesn't cover North America. Skipping.")
+                continue
+
+            if not slc_dataset["_source"]["metadata"].get("processing_mode") == "forward":
+                logging.info("dataset not captured in forward processing mode. Skipping.")
+                continue
+
+            if slc_dataset["_source"]["metadata"].get("intersects_north_america") \
+                    and slc_dataset["_source"]["metadata"].get("processing_mode") == "forward":
```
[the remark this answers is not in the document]
chrisjrd: the previous if statement functions as a guard clause. This one, while redundant, was meant to highlight or make more visible the conditions that trigger the body of work (code) inside it. I can remove it if you'd like, but i think it makes it easier to reason about the code when reading it, despite it being redundant.

data_subscriber/ionosphere_download.py
chrisjrd: i'll take another look. The code is meant to be accumulating and grouping the successes and failures, so if it isn't doing that, i'll need to fix it.

data_subscriber/ionosphere_download.py
chrisjrd: i'll take closer look

data_subscriber/ionosphere_download.py
chrisjrd: yes. this should be checking the local variable inside the loop. not the one outside. fixing.

I'm not sure about combining them. One block is meant to group the results, while the other one is directly part of the exception handling. There is a more concise way of writing the first block, but it escapes me what it is (similar in functionality to a dict comprehension, but may be more of a map or reduce).

i can combine them, but i wanted to keep them as distinct blocks since i felt they were performing different functions. Think about maybe if we extracted the blocks into their own functions. The functions would be more "pure" (one input to one output/action) if their functionalities were not combined.

yes, we can combine them since they share the conditional, but i'm looking at it in terms of grouping the statements by their purpose.

data_subscriber/ionosphere_download.py
```diff
@@ -0,0 +1,340 @@
...
+def get_arg_timerange(args):
+    start_date = args.start_date
+    end_date = args.end_date
+
+    download_timerange = DateTimeRange(start_date, end_date)
+    logger.info(f"{download_timerange=}")
+    return download_timerange
+
+
+@backoff.on_exception(backoff.expo, exception=Exception, max_tries=3, jitter=None, giveup=lambda e: isinstance(e, IonosphereFileNotFoundException))
```
collinss-jpl, line 273: Does this `giveup` clause need to be added to the url-only version of this function below?

collinss-jpl (approved):
Major issues have been addressed. I think this branch could be merged now.

data_subscriber/ionosphere_download.py
chrisjrd: no, but it will definitely close any gaps and prevent any issues from the asymmetry with the other function. fixing.

hhlee445 (approved):

hhlee445 (approved):

PR #557, opened 2023-07-20
Static layer toggle for historical processing

collinss-jpl:
This branch adds a feature to automatically disable static layer generation for the R2 PGE's whenever the processing mode is not set to "forward". This ensures that static layers are never generated when in "historical" or "reprocessing" mode.

This branch was tested by using the daac_data_subscriber.py script with the `--processing-mode=historical` option, and with static layers enabled for CSLC-S1 in settings.yaml. Within the `run_sciflo_L2_CSLC_S1.log` file for the job, the following should appear near the top of the log:

```
2023-07-20 21:38:19 opera_pcm:INFO [precondition_functions.py:765] Getting ENABLE_STATIC_LAYERS setting for PGE CSLC_S1
2023-07-20 21:38:19 opera_pcm:INFO [precondition_functions.py:773] Processing mode for L2_CSLC_S1 is set to historical, static layer generation will be DISABLED.
```

### DIFF
--- a/opera_chimera/constants/opera_chimera_const.py
+++ b/opera_chimera/constants/opera_chimera_const.py
@@ -20,6 +20,12 @@ class OperaChimeraConstants(ChimeraConstants):
 
     PROCESSING_MODE_KEY = "processing_mode"
 
+    PROCESSING_MODE_FORWARD = "forward"
+
+    PROCESSING_MODE_HISTORICAL = "historical"
+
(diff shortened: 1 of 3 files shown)


hhlee445 (commented):
@collinss-jpl We should allow the static layer generation with the reprocessing mode. The smoke test runs as reprocessing mode to ensure that we can generate all products.

collinss-jpl:
> 

Understood, will push a fix up shortly

PR #570, opened 2023-08-02
When downloading batch id=1, do a direct ES query for that granule id to save resources for HLS

hhlee445:


philipjyoon:
@hhlee445 this fix was only for R1 software. It will break R2 --batch-ids behavior as it's currently written. It's meant to merge into an R1 maintenance branch where only R1 code exists and is expected to deploy into existing R1 systems. Does such branch exist?

If we want to merge into develop anyway, I can add one more conditional so that it limits the behavior change to just R1 portion.

philipjyoon:
@hhlee445 I understand why we want to merge this into develop now. I will update the code so that only R1 business logic is affected by this change later today.

philipjyoon:
Will be implementing a different, more complete, solution on a different branch. Ignore this pull request and branch for now.

PR #583, opened 2023-08-16
Orbit time range start margin fix

collinss-jpl:
This branch updates stage_orbit_file.py to ensure the selected Orbit file has a start time that occurs at least 60 minutes prior to the start time of the corresponding SLC SAFE archive. This change is to bring the staging logic in line with what is expected by the s1_reader library utilized by the R2 SAS.

### DIFF
--- a/tests/tools/test_stage_orbit_file.py
+++ b/tests/tools/test_stage_orbit_file.py
@@ -97,7 +97,7 @@ class TestStageOrbitFile(unittest.TestCase):
             <opensearch:startIndex>0</opensearch:startIndex>
             <opensearch:itemsPerPage>10</opensearch:itemsPerPage>
             <entry>
-            <title>S1A_OPER_AUX_POEORB_OPOD_20220521T081912_V20220430T225942_20220502T005942</title>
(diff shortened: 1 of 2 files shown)


seongsujeong (commented):
I was the one who implemented the metadata correction mentioned in the email thread this morning.

I'm not familiar with the POEORB file name `S1A_OPER_AUX_POEORB_OPOD_20220521T081912_V20220430T200000_20220502T005942` in the unit test code `test_stage_orbit_file.py`, but the temporal margin that is applied in `stage_orbit_file.py` looks consistent with what we've changed in the point release last month.

collinss-jpl:
> I was the one who implemented the metadata correction mentioned in the email thread this morning.
> 
> I'm not familiar with the POEORB file name `S1A_OPER_AUX_POEORB_OPOD_20220521T081912_V20220430T200000_20220502T005942` in the unit test code `test_stage_orbit_file.py`, but the temporal margin that is applied in `stage_orbit_file.py` looks consistent with what we've changed in the point release last month.

Thanks for taking a look Seongsu, that POEORB file name is basically made up to satisfy that particular unit test.

hhlee445 (approved):

Issue #590, opened 2023-08-21
[Bug]: Network Timeouts and Lack of Retries in CMR Script


riverma:
### Checked for duplicates

Yes - I've already checked

### Describe the bug

While running the CMR audit script, we encountered a `504 Gateway Time-out` error when making network requests to the CMR server. The script may lack a robust retry mechanism to handle transient network errors, leading to potential failures. Additionally the script may just be making too many queries to the CMR server leading to the server to timeout our requests. 

Here's an example of the error message (full log [here](https://github.com/nasa/opera-sds-pcm/files/12402457/cmr_audit_log.txt)):
```
ERROR:  815321 backoff:_common.py:_log_giveup:120 - Giving up fetch_post_url(...) after 3 tries (aiohttp.client_exceptions.ClientResponseError: 504, message='Gateway Time-out', url=URL('https://cmr.earthdata.nasa.gov/search/granules.umm_json'))
Traceback (most recent call last):
  File "cmr_audit_hls.py", line 251, in <module>
    asyncio.run(run(sys.argv))
  File "/usr/local/Cellar/python@3.11/3.11.4_1/Frameworks/Python.framework/Versions/3.11/lib/python3.11/asyncio/runners.py", line 190, in run
    return runner.run(main)
           ^^^^^^^^^^^^^^^^
  File "/usr/local/Cellar/python@3.11/3.11.4_1/Frameworks/Python.framework/Versions/3.11/lib/python3.11/asyncio/runners.py", line 118, in run
    return self._loop.run_until_complete(task)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/Cellar/python@3.11/3.11.4_1/Frameworks/Python.framework/Versions/3.11/lib/python3.11/asyncio/base_events.py", line 653, in run_until_complete
    return future.result()
           ^^^^^^^^^^^^^^^
  File "cmr_audit_hls.py", line 193, in run
    cmr_granules_s30, cmr_granules_s30_details = await async_get_cmr_granules_hls_s30(temporal_date_start=cmr_start_dt_str, temporal_date_end=cmr_end_dt_str)
                                                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #595, opened 2023-08-23
Issue 590: CMR 504 timeout responses

chrisjrd:
## Purpose
- To introduce concurrency control in the `cmr_audit_hls.py` script by adding a semaphore
- To share more details about dependencies being installed within the README 
## Proposed Changes
- [CHANGE] Introduced a semaphore in `cmr_audit_hls.py` to control concurrent async tasks
- [CHANGE] Updated the pip installation command in `README.md` by adding the `-e` flag to enhance clarity and accuracy
## Issues
- #590 
## Testing
- The changes have been reviewed and tested locally without significant timeouts
- Further testing should be performed by OPS

### DIFF
--- a/tools/ops/README.md
+++ b/tools/ops/README.md
@@ -18,7 +18,7 @@ See `pcm_audit/*audit.py --help`, documentation comments, and source code for mo
 1. Create a python virtual environment.
     1. RECOMMENDED: move `pip.conf` into the resulting `venv/` directory.
 2. Activate the virtual environment and install the script dependencies referenced in the imports section as needed.
-    1. RECOMMENDED: install dependencies listed in the relevant section of `setup.py` using the following command `pip install '.[audit]'`
+    1. RECOMMENDED: install dependencies listed in the relevant section of `setup.py` using the following command `pip install -e '.[audit]'`
 
 ### Running locally
 
@@ -43,7 +43,7 @@ See `cmr_audit/cmr_audit.py --help`, documentation comments, and source code for
 1. Create a python virtual environment.
     1. RECOMMENDED: move `pip.conf` into the resulting `venv/` directory.
 2. Activate the virtual environment and install the script dependencies referenced in the imports section as needed.
-    1. RECOMMENDED: install dependencies listed in the relevant section of `setup.py` using the following command `pip install '.[cmr_audit]'`
+    1. RECOMMENDED: install dependencies listed in the relevant section of `setup.py` using the following command `pip install -e '.[cmr_audit]'`
 
 ### Running locally
 
--- a/tools/ops/cmr_audit/cmr_audit_hls.py
+++ b/tools/ops/cmr_audit/cmr_audit_hls.py
@@ -80,6 +80,7 @@ async def async_get_cmr_dswx(dswx_native_id_patterns: set):
 
     request_url = "https://cmr.earthdata.nasa.gov/search/granules.umm_json"
 
+    sem = asyncio.Semaphore(15)
     async with aiohttp.ClientSession() as session:
         post_cmr_tasks = []
         for i, dswx_native_id_pattern_batch in enumerate(dswx_native_id_pattern_batches, start=1):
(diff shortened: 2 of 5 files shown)


tools/ops/README.md
```diff
@@ -18,7 +18,7 @@ See `pcm_audit/*audit.py --help`, documentation comments, and source code for mo
 1. Create a python virtual environment.
     1. RECOMMENDED: move `pip.conf` into the resulting `venv/` directory.
 2. Activate the virtual environment and install the script dependencies referenced in the imports section as needed.
-    1. RECOMMENDED: install dependencies listed in the relevant section of `setup.py` using the following command `pip install '.[audit]'`
+    1. RECOMMENDED: install dependencies listed in the relevant section of `setup.py` using the following command `pip install -e '.[audit]'`
```
riverma, line 21: Hmm. Why do we want to use editable mode in a production setting?

tools/ops/cmr_audit/cmr_audit_hls.py
```diff
@@ -80,6 +80,7 @@ async def async_get_cmr_dswx(dswx_native_id_patterns: set):
 
     request_url = "https://cmr.earthdata.nasa.gov/search/granules.umm_json"
 
+    sem = asyncio.Semaphore(15)
```
riverma, line 83: This is great 💯. Any use of async network calls will benefit from semaphores like this. Do we expect the value of `15` to change at any point?

tools/ops/cmr_audit/cmr_audit_slc.py
```diff
@@ -102,6 +100,7 @@ async def async_get_cmr(
     request_url = "https://cmr.uat.earthdata.nasa.gov/search/granules.umm_json"
```
riverma, line 100: Remember to @hhlee445 @chrisjrd to swap to production CMR before including as part of a release.

tools/ops/cmr_audit/cmr_audit_utils.py
```diff
@@ -70,8 +71,8 @@ async def async_get_cmr_granules(collection_short_name, temporal_date_start: str
         logger.debug("Batching tasks")
         cmr_granules = set()
         cmr_granules_details = {}
-        task_chunks = list(more_itertools.chunked(post_cmr_tasks, 30))
-        for i, task_chunk in enumerate(task_chunks, start=1):  # CMR recommends 2-5 threads.
+        task_chunks = list(more_itertools.chunked(post_cmr_tasks, len(post_cmr_tasks)))  # CMR recommends 2-5 threads.
```
riverma, line 74: Do we still need the comment `# CMR recommends 2-5 threads.`? It's not clear where in the code the threads are limited to 2-5.

tools/ops/cmr_audit/cmr_client.py
```diff
@@ -1,56 +1,61 @@
...
+from math import ceil
 
 logger = logging.getLogger(__name__)
 
 
-async def async_cmr_post(url, data: str, session: aiohttp.ClientSession):
-    page_size = 2000  # default is 10, max is 2000
-    data += f"&page_size={page_size}"
+async def async_cmr_post(url, data: str, session: aiohttp.ClientSession, sem: Optional[asyncio.Semaphore]):
+    sem = sem if sem is not None else contextlib.nullcontext()
```
riverma, line 15: `else contextlib.nullcontext()` - very nice 👍

riverma (commented):
Great job here @chrisjrd. Nice use of the semaphore approach. Only suggestions are to add a bit of clarification for some comments in the code and also clarify the increase in max tries.

tools/ops/README.md
chrisjrd: it was a typo to not have the `-e`. But to support ad-hoc code changes, like updating constants, `-e` is easier for ops users to work with since it doesn't require re-running `pip install ...`: users are very likely to overlook this step.

tools/ops/cmr_audit/cmr_audit_hls.py
chrisjrd: I expected it to change rarely, if ever. If the concurrency level becomes unreliable, we'd dial it down in a subsequent update to the tool. It's not something that changes often enough to warrant making it a configurable item.

tools/ops/cmr_audit/cmr_client.py
chrisjrd: This is the recommended way to set up a conditional context manager (`with mycontext: ...`). It's basically a context manager with no-op __enter__() and __exit__() methods, but we don't have to write such a context manager ourselves, nor have a slightly more complicated if-else for the context/no-context cases.

tools/ops/cmr_audit/cmr_client.py
```diff
@@ -67,7 +72,7 @@ def giveup_cmr_requests(e):
 @backoff.on_exception(
     backoff.expo,
     exception=(aiohttp.ClientResponseError, aiohttp.ClientOSError),  # ClientOSError happens when connection is closed by peer
-    max_tries=3,
+    max_tries=7,
```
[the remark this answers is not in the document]
chrisjrd: that is a good idea. will implement.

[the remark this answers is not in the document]
riverma: Thanks.

riverma (approved):

hhlee445 (approved):

PR #597, opened 2023-08-27
Fix fix_get_slc_s1_dem() for antimeridian crossing

gshiroma:
This PR fixes the function fix_get_slc_s1_dem() for S1 datasets that cross the antimeridian.

When there S1 dataset crosses the antimeridian, the western and eastern longitude boundaries cannot be computed directly using `min()` and `max()` functions because the longitude coordinates are "wrapped" by `360` degrees. This PR fixes it by first detecting the antimeridian crossing using the statement `if lon_max - lon_min > 180:` and wrapping the negative longitude values up by `360` degrees: `lons = [lon + (lon < 0) * 360 for lon in lons]`. Afterwards, the `min()` and `max()` values can be computed normally.

### DIFF
--- a/opera_chimera/precondition_functions.py
+++ b/opera_chimera/precondition_functions.py
@@ -996,6 +996,12 @@ class OperaPreConditionFunctions(PreConditionFunctions):
         lon_min = min(lons)
         lon_max = max(lons)
 
+        # check for the antimeridian crossing:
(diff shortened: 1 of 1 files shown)


collinss-jpl:
This fix has been incorporated into #599 and will be merged with that PR

PR #603, opened 2023-09-04
Fix GDAL translate cropping coordinates to avoid the removal of the last pixel in X or Y directions

gshiroma:
This PR updates the cropping coordinates that are passed to GDAL translate to crop the DEM and avoid the removal of the last pixel of the dataset (DEM). 

This update is relevant to the antimeridian crossing case, in which the last (east-most) DEM pixels will be located in the middle of the staged DEM as longitude coordinates are usually wrapped around +/- 180.

GDAL translate snaps `projWin` coordinates to the map grid of the input dataset. 

~* We add `0.4` to the maximum length and width to make sure that GDAL Translate will not round down the ending coordinates and remove the last pixel in X or Y directions. The value `0.4` less than `0.5` to avoid that GDAL translate adds an extra pixel to the input extents. *~

UPDATE: We snap the minimum and maximum coordinates using the DEM grid. This is done using a lambda function called `snap_coord` and the snap value from the DEM pixel sizes `xres` and `yres` and offset from the starting coordinates `input_x_min` and `input_y_max`.

### DIFF
--- a/opera_chimera/precondition_functions.py
+++ b/opera_chimera/precondition_functions.py
@@ -996,12 +996,6 @@ class OperaPreConditionFunctions(PreConditionFunctions):
         lon_min = min(lons)
         lon_max = max(lons)
 
-        # check for the antimeridian crossing:
-        if lon_max - lon_min > 180:
-            lons = [lon + (lon < 0) * 360 for lon in lons]
-            lon_min = min(lons)
-            lon_max = max(lons)
-
         bbox = [lon_min, lat_min, lon_max, lat_max]  # WSEN order
 
         logger.info(f"Derived DEM bounding box: {bbox}")
--- a/opera_chimera/precondition_functions.py
+++ b/opera_chimera/precondition_functions.py
@@ -996,6 +996,12 @@ class OperaPreConditionFunctions(PreConditionFunctions):
         lon_min = min(lons)
         lon_max = max(lons)
 
+        # check for the antimeridian crossing:
+        if lon_max - lon_min > 180:
+            lons = [lon + (lon < 0) * 360 for lon in lons]
+            lon_min = min(lons)
+            lon_max = max(lons)
+
         bbox = [lon_min, lat_min, lon_max, lat_max]  # WSEN order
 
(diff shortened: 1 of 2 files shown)


collinss-jpl:
Hi @gshiroma, is this change meant to address the issues with RTC-S1 where a single burst out of the input granule fails the DEM coverage check, but others do not?

I've tested this change with several of the granules that exhibited this behavior, and they are still failing for the same burst.

gshiroma:
Thank you for testing it, @collinss-jpl . Did you test this change along with your changes from PR  #599? 
We expect to solve these antimeridian issues in two ways. One is through the fixes from these OPERA SDS/PCM PRs  (#603 and #599) and another is in the final RTC-S1 SAS (see updates [here](https://github.com/opera-adt/RTC/pull/69)). The SAS ancillary coverage check should be more robust and not fail if that last column of the DEM is missing.

gshiroma:
I was expecting the fixes here to solve the issues without the updates to the RTC-S1 SAS. What is the dataset that is still failing? I'm going to test it here. Thanks.

collinss-jpl:
These are the three that I tested:
S1B_IW_SLC__1SDV_20210730T183133_20210730T183145_028027_0357ED_AE8F
S1B_IW_SLC__1SDV_20210730T183014_20210730T183044_028027_0357ED_4B46
S1B_IW_SLC__1SDV_20210718T183132_20210718T183144_027852_0352CE_CDDA

gshiroma:
@collinss-jpl , I just updated this PR and the new code should fix all the errors related to this "missing pixel" at the eastern edge. It turns out GDAL translate rounds `xf` based on the `x0` that is provided. I sent you more info through slack. In order to get `xf` values rounded as expected, we need to snap `x0` coordinates. I updated the code to snap the minimum and maximum coordinates before we compare these coordinates with the DEM extents. I tested this code with the 11 "challenging antimeridian datasets" you sent me, and it worked correctly for all of them. I can provide you more info if you need.

collinss-jpl:
The relevant changes in this branch were incorporated into PR #599

PR #632, opened 2023-10-09
R2 PGE Static Layer Pipelines

### LINKED ISSUE #629: [New Feature]: Separate pipelines for RTC and CSLC Static layers generation
### Checked for duplicates

Yes - I've already checked

### Alternatives considered

No - I haven't considered

### Related problems

_No response_

### Describe the feature request

OPERA SDS wants to have separate pipelines for RTC and CSLC static layer generation. We can clone from RTC/CSLC standard product generation and enabled the static layer generation. Trigger rules for the  static layer generation will be off as default.

collinss-jpl:
## Purpose
- This branch refactors the workflows for the R2 PGEs to separate them into distinct SCIFLO jobs for the static layer products. Static layer product generation can now be enabled and disabled via trigger rule, rather than an update to settings.yaml. 
- Static layer products are now stored within their own subdirectory under products in S3 rolling storage.
- The static layer workflows can be run in PGE simulation mode
- The number of workers used for RTC static layer generation is now set to half the available cores. For the baseline workflow, this is still 3/4th the available cores. This is to help alleviate memory resource exhaustion when running the RTC static layer workflow on high-latitude SLC granules

## Issues
- Resolves #629 

## Testing
- This branch was tested on a dev cluster using the following granule in "reprocessing" mode: S1A_IW_SLC__1SDV_20220501T015035_20220501T015102_043011_0522A4_42CC 
- Both simulation and real PGE execution was tested, and in all cases both the baseline and static layer workflows completed successfully


opera_chimera/precondition_functions.py
```diff
@@ -241,6 +232,29 @@ def get_rtc_s1_num_workers(self):
...
+        Determines the number of workers/cores to assign to an RTC-S1-STATIC job
+        as a fraction of the total available.
+
+        """
+        logger.info(f"Evaluating precondition {inspect.currentframe().f_code.co_name}")
+
+        available_cores = os.cpu_count()
+
+        # Use 1/2 of the available cores for static layer processing
+        num_workers = max(int(round(available_cores / 2)), 1)
```
chrisjrd, line 246: here's the formula some other python library uses to specify a minimum number of threads to use.
`min(32, os.cpu_count() + 4)`. just for inspiration.

docker/hysds-io.json.SCIFLO_L2_CSLC_S1_hist
```diff
@@ -81,7 +81,7 @@
       "name": "container_working_dir",
       "from": "value",
       "type": "text",
-      "value": "/home/compass_user/scratch"
+      "value": "/home/compass_user"
```
chrisjrd, line 84: do we want to keep this change? seems out of place, so just checking that it is intentional.

chrisjrd (commented):
code changes look good. left minor comments. (no action required).
I still need to review the config files, if needed.

docker/hysds-io.json.SCIFLO_L2_CSLC_S1_hist
collinss-jpl: Yes, it was an intentional change since this value should match what is used for the non-historical version.

hhlee445 (approved):

PR #719, opened 2024-01-24
DISP-S1 PGE ER5.1 Integration

### LINKED ISSUE #708: [Task]: DISP-S1 PGE Engineering Release 5.0 Integration
### Checked for similar tasks

Yes - I've already checked

### Describe the task

Ticket to track integration of v3.0.0-er.5.0 of the DISP-S1 PGE with OPERA PCM.

As part of this integration, the existing workflow will be updated to integrate with the product metadata provided by the staging/trigger logic for DISP-S1, as well as add support for the retrieval of all ancillary input products.

### Prerequisites or Requirements

```bash
1.
2.
3.
...
```

### Expected Deliverables

```bash
1.
2.
3.
...
```

### Additional Context

_No response_

collinss-jpl:
## Purpose
- This branch integrates the DISP-S1 PGE v3.0.0-er.5.1 with the PCM framework. This release encapsulates the latest beta delivery of the DISP-S1 SAS.
- The DISP-S1 PGE can be invoked directly as an on-demand job from Tosca. The SCIFLO job is configured to use canned product metadata from S3, along with the "golden" input data from the small sample input frame. Execution should only take about 30 minutes with this small dataset.
- For this current integration, no ancillary inputs are staged or used by the DISP-S1 job, however, the SAS is still runnable in this state so it should be suitable for further integration testing.
- This update also enables dataset collection of any compressed CSLC products created by the PGE/SAS. Compressed CSLC products will be uploaded to S3 into the **Long Term Storage** bucket, under the `products/CSLC_S1_COMPRESSED` key. Within this location, compressed CSLC products will be grouped according to burst ID. 
  - An example URI: `s3://<LTS bucket>/products/CSLC_S1_COMPRESSED/compressed_t042_088905_iw1/compressed_t042_088905_iw1_20221119_20221213.h5`

## Issues
- Resolves #708 

## Testing
- This branch was tested using a dev cluster, using both simulation and real mode for the DISP-S1 PGE. Both the forward and historical versions of the SCIFLO job were executed. Note, however, that if both the forward and historical versions of the job are executed in simulation mode at the same time, no-clobber errors can occur since both versions produce identical sets of simulated products.


philipjyoon:
@collinss-jpl For Compressed CSLCs it would be better to use a deterministic naming scheme. This way the query code doesn't have to list all the files and try to find the right file - it can simply use the file name it knows it should be. Example is:
`s3://<LTS bucket>/products/CSLC_S1_COMPRESSED/compressed_t042_088905_iw1/compressed_t042_088905_iw1_a148.h5`

 We can use the `acquisition_cycle` instead of the actual datetime. It's calculated here: 
https://github.com/nasa/opera-sds-pcm/blob/develop/data_subscriber/url.py#L93

hhlee445 (approved):

PR #733, opened 2024-02-02
Bugfix 3.0.0 er.3.0 disp s1

philipjyoon:
Fixed race condition where forward query might pick up historical processing granules
Removed a sanity check that doesn't make sense for a lot of cases


philipjyoon:
Also created historical processing query test. All three (forward, reprocessing, historical) tests can be run simultaneously to test for performance and also look to look for any racing conditions.

hhlee445 (approved):

PR #749, opened 2024-02-25
Hls viirs update

hhlee445:
To test DSWx-HLS PGE execution with the HLS VIIRS data


hhlee445:
Will create a new pull request for HLS job specs only for release/3.0.0-rc.1.0

Issue #741, opened 2024-02-13
[Task]: Query/Download Jobs for CSLC-S1 Static Layer Products


collinss-jpl:
### Checked for similar tasks

Yes - I've already checked

### Describe the task

The DISP-S1 PGE will require CSLC-S1 Static Layer products as an ancillary input. Currently, there is no code in place to support performing a CMR query and download on the OPERA_L2_CSLC-S1-STATIC_V1 collection, so the goal of this ticket is to add such functionality. Integration of query/download logic into the DISP-S1 workflow will be handled in a separate ticket.

### Prerequisites or Requirements

```bash
1.
2.
3.
...
```

### Expected Deliverables

```bash
1.
2.
3.
...
```

collinss-jpl:
Closed by #751

PR #751, opened 2024-02-27
CSLC-S1 Static Layer Query/Download Support

collinss-jpl:
## Purpose
- This branch updates the daac_data_subscriber tool to support query and download of CSLC-S1 Static Layer products for use with the DISP-S1 workflow.
- Query of CSLC-S1 Static Layer products works as normal from the command line, just use "OPERA_L2_CSLC-S1-STATIC_V1" as the collection name.
- Download of static layer products is **not** currently supported via the command line. Static layer products are only downloaded as part of the standard CSLC-S1 product download workflow.

## Proposed Changes
- This branch also adds a significant amount of cleanup, refactoring and bug fixes to the data subscriber.
  - Argument parsing code has been moved into its own parser.py module to cut down on  SLOC in the main daac_data_subscriber.py script
  - Enumerations for commonly used string constants has been added to cmr.py and use of the new enumerations has been rolled out where appropriate
  - The run_download and run_query factory functions have been moved into daac_data_subscriber.py to fix circular dependency issues with imports
  - Several fixes made so CSLC-S1 query/download works as expected when using the "full" option
  - Several fixes made to cut down on noisy/redundant logging from modules outside of the data subscriber

## Issues
- Resovles #741 

## Testing
- Branch was tested on a dev cluster by using the daac_data_subscriber.py tool to query and download for all available collection types to try to ensure no regressions were introduced. PGE workflows were allowed to run in simulation mode.

### DIFF
--- a/data_subscriber/asf_cslc_download.py
+++ b/data_subscriber/asf_cslc_download.py
@@ -7,7 +7,12 @@ from os.path import basename
 from pathlib import PurePath, Path
 
 from data_subscriber import ionosphere_download
+from data_subscriber.cmr import Collection
+from data_subscriber.cslc.cslc_static_catalog import CSLCStaticProductCatalog
+from data_subscriber.download import SessionWithHeaderRedirection
+from data_subscriber.cslc_utils import parse_cslc_burst_id, build_cslc_static_native_ids
 from data_subscriber.asf_rtc_download import AsfDaacRtcDownload
+from data_subscriber.cslc.cslc_static_query import CslcStaticCmrQuery
 from util.aws_util import concurrent_s3_client_try_upload_file
 from util.conf_util import SettingsConf
 from util.job_submitter import try_submit_mozart_job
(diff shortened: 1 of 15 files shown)


philipjyoon:
@collinss-jpl Generally looks great and thanks for cleaning up some of the code.

1. `def parse_cslc_burst_id(native_id)` is a copy-paste of the first half of `def parse_cslc_burst_id(native_id)` Not a bid deal but would be cleaner to replace the first half of latter function to a call to the former so that we're not duplicating code.
2. I don't know much about CSLC Static usage CONOPs but is the datetime range of the those supposed to be identical to that of the CSLC product? CSLC querying is done by revision time and my understanding is that static products are temporally independently produced from the CSLC products.

hhlee445:
Verified that CSLC static layer products were listed under **InputFilePaths**

hhlee445 (approved):

PR #776, opened 2024-03-27
773 flexible s3 creds

philipjyoon:


data_subscriber/download.py
```diff
@@ -21,6 +21,7 @@
 
 logger = logging.getLogger(__name__)
 
+AWS_REGION = "us-west-2"
```
collinss-jpl, line 24: Not super important, but this exact key/value pair is written to .sds/config on deployed clusters, which I thought then becomes part of the envvars? Is there not a way we can access `AWS_REGION` from ~/.sds/config to avoid redefining it here?

collinss-jpl (approved):
Nice bit o'refactoring, love to see it!

hhlee445 (approved):

PR #797, opened 2024-04-10
Orbit file script backoff/retry fix

### LINKED ISSUE #796: [Bug]: Orbit File Script does not retry failed HTTP requests properly
### Checked for duplicates

Yes - I've already checked

### Describe the bug

All functions that make HTTP requests in the stage_orbit_file.py use the backoff decorator to automatically retry requests after an intermittent failure (such as request throttling by Orbit file server). The function that tests to see if the error is unrecoverable is not implemented correctly, so retries do not occur as expected for certain HTTP error codes.

### What did you expect?

When an HTTP request fails with error codes 401, 429, 500, 503, or 504, the backoff decorator retries the request every 15 seconds until the request succeeds or 5 minutes have elapsed.

### Reproducible steps

```bash
1.
2.
3.
...
```

### Environment

```bash
- Version of this software [e.g. vX.Y.Z]
- Operating System: [e.g. MacOSX with Docker Desktop vX.Y]
...
```

collinss-jpl:
## Purpose
- This PR fixes a bug in stage_orbit_file.py where the giveup function, which determines whether an automatic retry should occur or not, was not implemented correctly. This resulted in Orbit file downloads failing after only a single transient error was encountered, rather than once every ~15 seconds until success or 5 minutes have elapsed.

## Issues
- Fixes #796 

## Testing
- A unit test has been added to test_stage_orbit_file.py which simulates all the transient error codes we check for during an orbit file download, and ensures that the download is retried for each failure code.

### DIFF
--- a/tests/tools/test_stage_orbit_file.py
+++ b/tests/tools/test_stage_orbit_file.py
@@ -178,6 +178,8 @@ class TestStageOrbitFile(unittest.TestCase):
         response_429.status_code = 429
         response_500 = Response()
         response_500.status_code = 500
+        response_502 = Response()
+        response_502.status_code = 502
         response_503 = Response()
         response_503.status_code = 503
         response_504 = Response()
@@ -188,7 +190,7 @@ class TestStageOrbitFile(unittest.TestCase):
         response_200.status_code = 200
         response_200.raw = BytesIO(b'orbit file contents')
 
-        responses = [response_401, response_429, response_500, response_503, response_504, response_200]
+        responses = [response_401, response_429, response_500, response_502, response_503, response_504, response_200]
 
         # Set up a Mock function for session.get which will cycle through all
         # transient error codes before finally returning success (200)
--- a/tests/tools/test_stage_orbit_file.py
+++ b/tests/tools/test_stage_orbit_file.py
@@ -1,15 +1,29 @@
 #!/usr/bin/env python3
 
 import json
+import os.path
+import tempfile
 import unittest
+from io import BytesIO
+from unittest.mock import MagicMock, patch
+
+from requests import Response, Session
 
 import tools.stage_orbit_file
 from tools.stage_orbit_file import (ORBIT_TYPE_POE,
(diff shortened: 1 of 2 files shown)


tests/tools/test_stage_orbit_file.py
```diff
@@ -154,3 +168,45 @@ def test_select_orbit_file(self):
...
+
+        # Create the successful response
+        response_200.status_code = 200
+        response_200.raw = BytesIO(b'orbit file contents')
+
+        responses = [response_401, response_429, response_500, response_503, response_504, response_200]
+
+        # Set up a Mock function for session.get which will cycle through all
+        # transient error codes before finally returning success (200)
+        mock_requests_get = MagicMock(side_effect=responses)
```
philipjyoon, line 195: I'm not overly familiar with MagicMock but shouldn't the side_effects be list of Exceptions not Responses?

collinss-jpl: All HTTP requests made by stage_orbit_file.py have their result checked by [raise_for_status](https://www.geeksforgeeks.org/response-raise_for_status-python-requests/), which is what actually raises the exception caught by the backoff/retry decorator. So it is sufficient to only mock the `Response` objects returned by the request function itself (in this case `requests.get`)

hhlee445:
Executed historical processing with 35 max ASG.
Submitted 6237 slc_download jobs and 16 failed with the errors as 
_requests.exceptions.HTTPError: 502 Server Error: Bad Gateway for url: https://sentinel1.asf.alaska.edu/s3credentials_ 
Didn't get any orbit file download error.

collinss-jpl:
> Executed historical processing with 35 max ASG. Submitted 6237 slc_download jobs and 16 failed with the errors as _requests.exceptions.HTTPError: 502 Server Error: Bad Gateway for url: https://sentinel1.asf.alaska.edu/s3credentials_ Didn't get any orbit file download error.

I will add 502 to the list of error codes that we retry for so we hopefully see less of this type of failure going forward

PR #869, opened 2024-06-06
861 disp s1 use asf s3

philipjyoon:
For CSLC burst and static file downloads if the protocol is not `https`, the download code will use the ASF S3 paths as the local storage and bypass the download. All other download functionality, including CSLC https download, have been preserved.

CSLC ionosphere files are only available via http so those are still downloaded.

This code has been tested with HLS, SLC, RTC, and CSLC S3 and https download/SCIFLO and all are working correctly.

### DIFF
--- a/data_subscriber/asf_cslc_download.py
+++ b/data_subscriber/asf_cslc_download.py
@@ -4,6 +4,7 @@ import os
 
 from os.path import basename
 from pathlib import PurePath, Path
+import urllib.parse
 from datetime import datetime, timezone
 import boto3
 
@@ -64,7 +65,6 @@ class AsfDaacCslcDownload(AsfDaacRtcDownload):
                 )
                 logger.info(f"Uploading CSLC input files to S3")
                 cslc_files_to_upload = [fp for fp_set in cslc_products_to_filepaths.values() for fp in fp_set]
-                # TODO: uncomment this
                 cslc_s3paths.extend(concurrent_s3_client_try_upload_file(bucket=settings["DATASET_BUCKET"],
                                                                          key_prefix=f"tmp/disp_s1/{batch_id}",
                                                                          files=cslc_files_to_upload))
@@ -85,11 +85,18 @@ class AsfDaacCslcDownload(AsfDaacRtcDownload):
                 for p in cslc_s3paths:
                     # Split the following into bucket name and key
                     # 's3://asf-cumulus-prod-opera-products/OPERA_L2_CSLC-S1/OPERA_L2_CSLC-S1_T122-260026-IW3_20231214T011435Z_20231215T075814Z_S1A_VV_v1.0/OPERA_L2_CSLC-S1_T122-260026-IW3_20231214T011435Z_20231215T075814Z_S1A_VV_v1.0.h5'
-                    bucket = p.split("/")[2]
-                    key = "/".join(p.split("/")[3:])
-
+                    parsed_url = urllib.parse.urlparse(p)
+                    bucket = parsed_url.netloc
+                    key = parsed_url.path[1:]
                     granule_id = p.split("/")[-1]
-                    file_size = int(boto3.client("s3").head_object(Bucket=bucket, Key=key)["ContentLength"])
+
+                    try:
+                        head_object = boto3.client("s3").head_object(Bucket=bucket, Key=key)
+                    except Exception as e:
+                        logger.error("Failed when accessing the S3 object:" + p)
+                        raise e
+                    file_size = int(head_object["ContentLength"])
+
                     granule_sizes.append((granule_id, file_size))
 
                 cslc_files_to_upload = [Path(p) for p in cslc_s3paths] # Need this for querying static CSLCs
--- a/data_subscriber/asf_cslc_download.py
+++ b/data_subscriber/asf_cslc_download.py
@@ -56,17 +56,25 @@ class AsfDaacCslcDownload(AsfDaacRtcDownload):
 
             new_args.batch_ids = [batch_id]
 
-            # First, download the files from ASF
-            cslc_products_to_filepaths: dict[str, set[Path]] = await super().run_download(
-                new_args, token, es_conn, netloc, username, password, job_id, rm_downloads_dir=False
-            )
-
-            logger.info(f"Uploading CSLC input files to S3")
-            cslc_files_to_upload = [fp for fp_set in cslc_products_to_filepaths.values() for fp in fp_set]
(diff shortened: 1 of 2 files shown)


philipjyoon:
@collinss-jpl Minding taking a quick look at the static download portion of this code change? Pretty simple and I've tested it but just good to have a second set of eyes.

data_subscriber/asf_cslc_download.py
```diff
@@ -55,24 +55,50 @@ async def run_download(self, args, token, es_conn, netloc, username, password, j
...
-                                                                     key_prefix=f"tmp/disp_s1/{batch_id}",
-                                                                     files=cslc_files_to_upload))
+            # Download the files from ASF only if the transfer protocol is HTTPS
+            if args.transfer_protocol == "https":
+                cslc_products_to_filepaths: dict[str, set[Path]] = await super().run_download(
+                    new_args, token, es_conn, netloc, username, password, job_id, rm_downloads_dir=False
+                )
+                logger.info(f"Uploading CSLC input files to S3")
+                cslc_files_to_upload = [fp for fp_set in cslc_products_to_filepaths.values() for fp in fp_set]
+                # TODO: uncomment this
```
collinss-jpl, line 67: Could this TODO be removed now? One less to worry about...

data_subscriber/asf_cslc_download.py
```diff
@@ -55,24 +55,50 @@ async def run_download(self, args, token, es_conn, netloc, username, password, j
...
+                                                                         files=cslc_files_to_upload))
+
+                for granule_id, fp_set in cslc_products_to_filepaths.items():
+                    filepath = list(fp_set)[0]
+                    file_size = os.path.getsize(filepath)
+                    granule_sizes.append((granule_id, file_size))
+
+            # For s3 we can use the files directly so simply copy over the paths
+            else: # s3 or auto
+                logger.info("Skipping download CSLC bursts and instead using ASF S3 paths for direct SCIFLO PGE ingestion")
```
collinss-jpl, line 79: "Skipping download **of** CSLC bursts..."

Better yet, just shorten to "Using ASF S3 paths for direct SCIFLO PGE ingestion"

data_subscriber/asf_cslc_download.py
```diff
@@ -55,24 +55,50 @@ async def run_download(self, args, token, es_conn, netloc, username, password, j
...
+                if len(cslc_s3paths) == 0:
+                    raise Exception(f"No s3_path found for {batch_id}. You probably should specify https transfer protocol.")
+
+                for p in cslc_s3paths:
+                    # Split the following into bucket name and key
+                    # 's3://asf-cumulus-prod-opera-products/OPERA_L2_CSLC-S1/OPERA_L2_CSLC-S1_T122-260026-IW3_20231214T011435Z_20231215T075814Z_S1A_VV_v1.0/OPERA_L2_CSLC-S1_T122-260026-IW3_20231214T011435Z_20231215T075814Z_S1A_VV_v1.0.h5'
+                    bucket = p.split("/")[2]
+                    key = "/".join(p.split("/")[3:])
+
+                    granule_id = p.split("/")[-1]
```
collinss-jpl, line 91: I'm sure this code is doing what is intended, but its unintuitive to look at. Might be worth investigating if [urllib.parse](https://docs.python.org/3/library/urllib.parse.html#module-urllib.parse) can provide a cleaner interface than `split`

data_subscriber/asf_cslc_download.py
```diff
@@ -55,24 +55,50 @@ async def run_download(self, args, token, es_conn, netloc, username, password, j
...
+                    file_size = int(boto3.client("s3").head_object(Bucket=bucket, Key=key)["ContentLength"])
```
collinss-jpl, line 92: Though this should rarely be the case, we should have some error handling for the `head_object` call in case of sporadic S3 unavailability (or ASF just gives us an incorrect URL!)

Something like:
```python
try:
    head_object = boto3.client("s3").head_object(Bucket=bucket, Key=key)
except except botocore.exceptions.ClientError as e:
    <log details of error>

file_size = int(head_object["ContentLength"])
```

data_subscriber/asf_cslc_download.py
```diff
@@ -84,20 +110,35 @@ async def run_download(self, args, token, es_conn, netloc, username, password, j
...
+                )
+
+                logger.info("Uploading CSLC Static input files to S3")
+                cslc_static_files_to_upload = [fp for fp_set in cslc_static_products_to_filepaths.values() for fp in fp_set]
+                cslc_static_s3paths.extend(concurrent_s3_client_try_upload_file(bucket=settings["DATASET_BUCKET"],
+                                                                                key_prefix=f"tmp/disp_s1/{batch_id}",
+                                                                                files=cslc_static_files_to_upload))
+            # For s3 we can use the files directly so simply copy over the paths
+            else:  # s3 or auto
+                logger.info("Skipping download CSLC static files and instead using ASF S3 paths for direct SCIFLO PGE ingestion")
```
collinss-jpl, line 128: Same comment as above

collinss-jpl (approved):
I see nothing wrong with the implementation, just had a few "nice-to-have" findings. Thanks @philipjyoon!

data_subscriber/asf_cslc_download.py
philipjyoon: Good catch

data_subscriber/asf_cslc_download.py
philipjyoon: Agreed but I wanted to put the "skipping" verbiage there because a little above on line 51 it says that it's "downloading" files. Wanted to contrast against that.

data_subscriber/asf_cslc_download.py
philipjyoon: Same as above

data_subscriber/asf_cslc_download.py
philipjyoon: Agreed. If the boto3 call fails then the whole job should fail anyway (need that S3 file to process DISP-S1) but we can at least write out a sensical error.

data_subscriber/asf_cslc_download.py
philipjyoon: Great idea. Didn't know about [urllib.parse](https://docs.python.org/3/library/urllib.parse.html#module-urllib.parse)

philipjyoon:
All comments resolved and ready for merge

hhlee445 (approved):

Issue #896, opened 2024-07-02
[Bug]: DISP-S1 historical query job should submit download job w processing mode


philipjyoon:
### Checked for duplicates

Yes - I've already checked

### Describe the bug

The download job uses processing_mode to determine whether or not to pass that information to PGE. This information goes further into runconfig and that affects the SAS behavior. Two changes needed:
1. Download jobs must accept --processing-mode as a parameter. It does not currently. The usual daac...subscriber and docker files (for CSLC only) must be changed.
2. CSLC query job needs to pass processing_mode when submitting download jobs.

### What did you expect?

n/t

### Reproducible steps

```bash
1.
2.
3.
...
```

### Environment

```bash
- Version of this software [e.g. vX.Y.Z]
- Operating System: [e.g. MacOSX with Docker Desktop vX.Y]
...
```

philipjyoon:
Query job already passes --processing-mode parameter when submitting download jobs

PR #897, opened 2024-07-02
#896: CSLC download job now accepts --processing-mode as a parameter

philipjyoon:
## Purpose
CSLC download job now accepts --processing-mode as a parameter. Query job already passes --processing-mode parameter when submitting download jobs so it did not require changing.

### DIFF
--- a/docker/hysds-io.json.cslc_download
+++ b/docker/hysds-io.json.cslc_download
@@ -47,7 +47,7 @@
       "optional": true
     },
     {
-      "name": "processing_mode",
+      "name": "proc_mode",
       "from": "submitter",
       "placeholder": "e.g. --processing-mode=forward",
       "optional": true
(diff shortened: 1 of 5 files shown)


docker/hysds-io.json.cslc_download
```diff
@@ -46,6 +46,12 @@
       "placeholder": "e.g. --use-temporal",
       "optional": true
     },
+    {
+      "name": "processing_mode",
+      "from": "submitter",
+      "placeholder": "e.g. --processing-mode",
```
chrisjrd, line 52: the example placeholder seems wrong. i would have imagined it to be like `--processing-mode=forward` or something like that.

chrisjrd (approved):
left a minor comment. otherwise, looks good to me.

hhlee445 (approved):

PR #959, opened 2024-08-21
DSWx-S1 Validator Updates

riverma:
## Purpose
- Improvements and bug fixes to ensure the validator functions as expected for OPS and I&T use cases
## Proposed Changes
- [CHANGE] UTC timestamp support for entering dates with a "Z" at the end
- [CHANGE] Fixed bug to include logic for gathering DSWx-S1 granules to support > 1000 use cases
- [CHANGE] Logic to ensure comparison of missing input RTC bursts includes only bursts that are missing for the given tile set, not all tile sets. This allows the validator to detect if we processed the wrong tile in cases where RTC bursts were used for multiple tile sets.

## Issues
- N/A
## Testing
- Tested locally

### DIFF
--- a/report/dswx-s1-validator/dswx_s1_validator.py
+++ b/report/dswx-s1-validator/dswx_s1_validator.py
@@ -225,7 +225,7 @@ def generate_url_params(start, end, endpoint = 'OPS', provider = 'ASF', short_na
     }
 
     # Set CMR param to ignore granule searches prior to a certain date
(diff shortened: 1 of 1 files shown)


sjlewis-jpl (commented):
I don't see any critical changes being needed here, so I'll leave this as a "Comment" review.  I made a handful of inline comments with some questions and suggestions.

hhlee445 (approved):

hhlee445 (approved):

PR #968, opened 2024-08-29
DISP-S1 Workflow ECMWF Usage Update

### LINKED ISSUE #967: [Task]: Implement ECMWF file usage in DISP-S1 Workflow
### Checked for similar tasks

Yes - I've already checked

### Describe the task

Ticket to implement inclusion of ECMWF files when staging products for use with a DISP-S1 job.

This feature should include a toggle (in settings.yaml) to bypass inclusion of ECMWF files for ease of testing while we are waiting for files to be populated in S3.

The staging logic should also query S3 to ensure existence of the expected files (based on acquisition dates of input CSLC products) before they are included with the RunConfig for the job.

### Prerequisites or Requirements

```bash
1.
2.
3.
...
```

### Expected Deliverables

```bash
1.
2.
3.
...
```

### Additional Context

_No response_

collinss-jpl:
## Purpose
- This branch represents an initial integration of support for ECMWF (Troposphere) file usage with the DISP-S1 PGE workflow. A precondition function in the Chimera workflow for DISP-S1 will now derive the set of ECMWF files for use with a DISP-S1 job based on the CSLC products (compressed and standard) staged for the job. Note that further refinements may be required to this branch once a sufficient number of subset/compressed ECMWF files are available for further testing.
- A flag to control the behavior of the precondition function has also been added to settings.yaml, named `STRICT_ANCILLARY_USAGE`. If this flag is set to True, then any expected ECMWF files that are not available in S3 will result in a failed DISP-S1 job. If set to False, then any missing files will result in a DISP-S1 job with **no troposphere files provided**. This was added to allow testing to continue while the back catalog of subset and compressed ECMWF files is still being populated. When in production, the `STRICT_ANICLLARY_USAGE` flag should be set to True.
- Lastly, a new utility module, `ecmwf_utils.py`, has been added to collect the functions used to derive ECMWF S3 paths, and check for their availability. One function in particular, `find_ecmwf_for_datetime()`, has been added for use with the CSLC file staging logic to ensure that all ECMWF files are available before triggering a DISP-S1 job.

## Issues
- Resolves #967 

## Testing
- Unit tests have been added for both the new `ecmwf_utils.py` module, as well as the new precondition function used to assign ECMWF files to the DISP-S1 RunConfig.
- This branch was also tested on a dev cluster using both simulation and real mode for DISP-S1, as well as with the `STRICT_ANCILLARY_USAGE` flag enabled and disabled.

### DIFF
--- a/conf/settings.yaml
+++ b/conf/settings.yaml
@@ -154,6 +154,11 @@ DISP_S1:
   # Amount of margin in km to apply to staged ancillaries (DEM)
   ANCILLARY_MARGIN: 50
 
+  # Set flag this to False to allow a DISP-S1 job to continue if certain
+  # ancillaries are not available (such as troposphere files). If True, any
+  # missing files will result in a RuntimeError, failing the job.
+  STRICT_ANCILLARY_USAGE: !!bool false
+
 DSWX_NI:
   # Amount of margin in km to apply to staged ancillaries (DEM/Worldcover)
   ANCILLARY_MARGIN: 50
--- a/data_subscriber/cslc_utils.py
+++ b/data_subscriber/cslc_utils.py
@@ -167,6 +167,20 @@ def parse_cslc_file_name(native_id):
     acquisition_dts = match_product_id.group("acquisition_ts")  # e.g. 20210705T183117Z
     return burst_id, acquisition_dts
 
+
+def parse_compressed_cslc_file_name(native_id):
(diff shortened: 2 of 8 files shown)


collinss-jpl:
ECMWF file usage will be handled in a separate PGE from DISP-S1, so this PR is not needed at this time.

PR #1059, opened 2025-01-16
Data Subscriber RTC Download Code Refactor

collinss-jpl:
## Purpose
- This branch introduces some light refactoring of the code used for RTC-S1 product download and DSWx-S1 job triggering. The primary goal is to consolidate all RTC download code to the `AsfDaacRtcDownload` to make the class more suitable for future extension and reuse when download/triggering is implemented for DSWx-NI.

## Proposed Changes
- The `run_rtc_download` function (and associated helper functions) within `daac_data_subscriber.py` has been refactored into the `run_download` implementation within the `AsfDaacRtcDownload` class. This also simplifies the logic of the factory functions within `daac_data_subscriber.py`
- The superclass call to `DaacDownload.run_download()` in `AsfDaacCslcDownload.run_download()` has been modified to bypass the version now implemented in `AsfDaacRtcDownload`
- The `--file` argument has been removed from the `daac_data_subscriber.py` argparser since it was unused.
- To further simplify the logic within the factory functions in `daac_data_subscriber.py`, a `run_query` method has been added to `RtcCmrQuery` that wraps the asynchronous version (now called `run_query_async`)

## Testing
- Branch has been deployed and tested on development cluster with all PGEs running in simulation mode.
- Updates to download logic has been tested by enabling the download triggers and letting the cluster run for about ~8 hours
- Downloads have also been tested by submitting a few ad-hoc queries directly from mozart

### DIFF
--- a/data_subscriber/asf_cslc_download.py
+++ b/data_subscriber/asf_cslc_download.py
@@ -70,7 +70,8 @@ class AsfDaacCslcDownload(AsfDaacRtcDownload):
 
             # Download the files from ASF only if the transfer protocol is HTTPS
             if args.transfer_protocol == "https":
-                cslc_products_to_filepaths: dict[str, set[Path]] = super().run_download(
+                # Need to skip over AsfDaacRtcDownload.run_download() and invoke base DaacDownload.run_download()
+                cslc_products_to_filepaths: dict[str, set[Path]] = super(AsfDaacRtcDownload, self).run_download(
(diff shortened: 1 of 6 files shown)


philipjyoon:
Looks good to me. To be clear:
1. When the testing was run for ~8 hrs were the PGEs in simulation mode or non-sim mode?
2. If in simulation mode, does that still effectively validate (explicit check or consumption) of all ancillary input files that are downloaded by RTCDownload? (if any)

hhlee445 (approved):

PR #1061, opened 2025-01-16
1056 download from daac to s3

philipjyoon:
Refilsl all the lost DISP-S1 products from ASF back into the S3 bucket. Performs the following:

- Take in a text file w list of frame IDs
- Perform CMR query (assuming these are being cataloged to CMR) to find all products for each frame ID from ASF-UAT
- Compare to see if they exist in the S3 already
- If they don't exist, then copy the file over

This has been tested for several frames from ASF UAT to INT-POP1 S3 and the PST confirmed that it worked correctly.

Reuses some code from the opera_validator and it had to be refactored a bit to be useful.


tools/download_from_daac.py
```diff
@@ -0,0 +1,111 @@
...
+        print(f"Searching for DISP-S1 products with native-id pattern: {native_id_pattern}")
+    extra_params = {"native-id[]": native_id_pattern, "options[native-id][pattern]": "true"}
+
+    # Retrieve all DISP-S1 products from CMR within the acquisition time range as a list of granuleIDs
+    all_disp_s1 = retrieve_r3_products(smallest_date, greatest_date, "UAT", _DISP_S1_PRODUCT_TYPE, extra_params=extra_params)
+    for disp_s1 in all_disp_s1:
+
+        # NOTE: This time filter isn't really used - we hard-coded it to 1999 to 2099 for now. If at some point we want to
+        # filter by time as well, we will need to configure this a bit to make it work
+        for attrib in disp_s1.get("umm").get("AdditionalAttributes"):
```
collinss-jpl, line 53: I don't understand why this loop over "AdditionalAttributes" is needed. `attrib` doesn't seem to be used anywhere else.

collinss-jpl (approved):
Code looks good, just had one question about a loop that maybe isn't needed?

PR #1066, opened 2025-01-27
1062 disp s1 validator inputs

philipjyoon:
DISP-S1 validation now validates both ways: 1) DISP-S1 products for whose input CSLC are not known and 2) CSLC input files whose DISP-S1 products have not been created.

In addition, several edge-cases are now handled:
1. Non-k-complete CSLC input files are no longer determined to be failing validation. They become don't-cares.
2. DISP-S1 products that are within the non-k-complete timeframes are ignored.
3. Filter out input CSLC files that are not within the consistent database when validating in historical mode.
4. `--view_all_cslc_groupings` option show validation groupings for all CSLC input files.


hhlee445:
@philipjyoon is there a wiki page for usage of DISP-S1 validation script?

Issue #1077, opened 2025-02-06
[New Feature]: Renable tasks_acks_late flag for celery job worker


philipjyoon:
### Checked for duplicates

Yes - I've already checked

### Alternatives considered

Yes - and alternatives don't suffice

### Describe the feature request

This ticket we want to actually reverse what we did in https://github.com/nasa/opera-sds-pcm/issues/1053 It turns out there are a couple major unintended consequences (one of them being almost a bug in celery implementation and non-compliance in rabbitmq) and until we get around those consequences, late_ack=False slows down processing too much. And it also specifically affects DISP-S1 SCIFLO negatively the most which is THE PGE whose performance we were trying to improve.

In that ticket we created a new celery config file for asg verdi workers. So we would just set `task_acks_late = True`
https://github.com/nasa/opera-sds-pcm/commit/f24305d2349bc33275ea24c4ae6a22ab0bc62194

Issue #1076, opened 2025-02-05
[New Feature]: Make number of processes used DISP-S1 SCIFLO be settings.yaml-configurable


philipjyoon:
### Checked for duplicates

Yes - I've already checked

### Alternatives considered

Yes - and alternatives don't suffice

### Describe the feature request

We currently use a hard-coded formula for determining how many processes to use in DISP-S1 SCIFLO processing. We should make this formula a bit more flexible by getting some of the parameters from `settings.yaml` so that change to this formula doesn't require a new PCM release and deployment. It's currently here:
https://github.com/nasa/opera-sds-pcm/blob/develop/opera_chimera/precondition_functions.py#L385
`n_parallel_bursts = max(int(round(available_cores / 2)) + 1, 1)`

Make the following changes to that function:
1. Parameterize the division factor and the constant factor so that they are retrieved from `settings.yaml`. Have reasonable values if they are not found.
2. Make the default value of the denominator be 4 instead of 2. The constant factor can still be 1.
3. I'd also make change that formula so that it's a multiplication instead of division. This allows for maximum flexibility and the most straightforward. So the currently formula would look:  `n_parallel_bursts = max(int(round(available_cores * 0.5 )) + 1, 1)`
4. At this point might as well parameterize `theads_per_worker` too
https://github.com/nasa/opera-sds-pcm/blob/develop/opera_chimera/precondition_functions.py#L380
`threads_per_worker = available_cores`

PR #1078, opened 2025-02-06
1076 1077 disp s1 sciflo num processes and asg celery late ack

philipjyoon:
## Purpose
- This PR addresses two tickets at once: 1076 and 1077. I put them into one branch for quicker testing because time is of essence for these changes today.
## Proposed Changes
- Change back `task_acks_late` to be `True`
- Parameterize num process determination for DISP-S1 SCIFLO PGE

## Issues
- https://github.com/nasa/opera-sds-pcm/issues/1076
- https://github.com/nasa/opera-sds-pcm/issues/1077
## Testing
- I ran DISP-S1 with n_procs = 9 instead of 17 we'd been using. Idea is to reduce the number of processes to reduce memory usage so that we are not thrashing on swap memory. I wanted to make sure that taking that approach wouldn't slow us down. It actually seems to have improved running time on average by 5%. There is some randomness to it but there is overall trend. Run time is in seconds. The run-time also encompasses ALL machine activity including downloading all input files and copying them out to S3 which happen outside of the SAS container.
- This should basically be an apples-to-apples comparison. Used the exact same EC2 type and ran through the exact same frames and sensing datetime range. These were the first k=15 set for each frame so theoretically these should be one of the faster runs for that frame.
- Also verified that late_ack behavior is what we expected and all jobs were being scheduled to run as opposed to being in unacked state and queued.
- Verified that the values in `settings.yaml` were being logged and used by DISP-S1 SCIFLO PGE

image

### DIFF
--- a/opera_chimera/precondition_functions.py
+++ b/opera_chimera/precondition_functions.py
@@ -376,10 +376,8 @@ class OperaPreConditionFunctions(PreConditionFunctions):
 
         available_cores = os.cpu_count()
 
-        # Use all available cores for threads_per_worker
-
         try:
-            threads_per_worker = self._settings.get("DISP_S1_NUM_THREADS")
+            threads_per_worker = self._settings["DISP_S1_NUM_THREADS"]
         except:
             threads_per_worker = available_cores
             logger.error(f"DISP_S1_NUM_THREADS not found in settings.yaml. Using default {threads_per_worker=}")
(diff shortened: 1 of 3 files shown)


opera_chimera/precondition_functions.py
```diff
@@ -377,13 +377,26 @@ def get_disp_s1_num_workers(self):
         available_cores = os.cpu_count()
 
         # Use all available cores for threads_per_worker
-        threads_per_worker = available_cores
 
-        logger.info(f"Allocating {threads_per_worker=} out of {available_cores} available")
+        try:
+            threads_per_worker = self._settings.get("DISP_S1_NUM_THREADS")
+        except:
```
collinss-jpl, line 383: This will not work as you expect, since you are using `get`, if "DISP_S1_NUM_THREADS" is not in settings.yaml, `threads_per_worker` will default to `None`.

I think you want to use direct dictionary indexing (`self._settings["DISP_S1_NUM_THREADS"]`) and then you can catch the `KeyError` in your except clause

collinss-jpl: Looks good, but if a key isn't found in settings.yaml, I would prefer it log to warning level rather than error, since its something we recover from by using default values.

philipjyoon: I can go either way. Changed to `warning`

collinss-jpl (approved):

philipjyoon:
Updated the PR with testing result. This PR is ready to be merged @hhlee445

Issue #1086, opened 2025-02-13
[Bug]: Allow for explicit provider in daac_data_subscriber.py query jobs


philipjyoon:
### Checked for duplicates

Yes - I've already checked

### Describe the bug

Current daac_data_subscriber maps collections (such as HLSS30) to provider (such as LPCLOUD) and does allow the user to specify an explicit provider. This is an issue when querying from a UAT CMR that uses a different provider than PROD, for the exact same collection. It also doesn't allow any provider that's not within its list of providers to be passed in.

### What did you expect?

n/t

### Reproducible steps

```bash
1.
2.
3.
...
```

### Environment

```bash
- Version of this software [e.g. vX.Y.Z]
- Operating System: [e.g. MacOSX with Docker Desktop vX.Y]
...
```

PR #1087, opened 2025-02-13
#1086: Added LPCLOUDUAT to the list of providers. Use explicitly pass…

philipjyoon:
…-in provider instead of mapping

Tested on INT cluster and works for both UAT and non-UAT CMR queries on HLS data.

### DIFF
--- a/data_subscriber/cmr.py
+++ b/data_subscriber/cmr.py
@@ -37,6 +37,7 @@ class Endpoint(str, Enum):
 
 class Provider(str, Enum):
     LPCLOUD = "LPCLOUD"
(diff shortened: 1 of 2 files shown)


collinss-jpl (approved):

philipjyoon:
Actually, this isn't complete. It looks like `--provider` is not an option in `query` mode. It's an option in `survey`. Need to add provider to query as well for this to work.

philipjyoon:
@hhlee445 This PR is ready to be merged. Tested in query mode all the way through.

hhlee445 (approved):

Issue #1102, opened 2025-03-06
[Bug]: CSLC download for forward/reprocessing error


philipjyoon:
### Checked for duplicates

Yes - I've already checked

### Describe the bug

`python3 ~/mozart/ops/opera-pcm/data_subscriber/daac_data_subscriber.py query     --collection-shortname=OPERA_L2_CSLC-S1_V1     --endpoint=OPS     --k=4     --m=1     --job-queue=opera-job_worker-cslc_data_download     --chunk-size=1     --processing-mode=reprocessing     --transfer-protocol=auto     --native-id=OPERA_L2_CSLC-S1_T093-197858-IW3_20231118T013640Z_20231119T073552Z_S1A_VV_v1.0`

The corresponding job dies w the following error:
```
[2025-03-06 20:02:18,572: INFO/get_downloads] Got len(downloads)=0 cslc granules downloads for batch_id=f24733_a2676
Traceback (most recent call last):
  File "/export/home/hysdsops/mozart/ops/opera-pcm/data_subscriber/daac_data_subscriber.py", line 144, in <module>
    main()
  File "/export/home/hysdsops/mozart/ops/opera-pcm/util/exec_util.py", line 35, in wrapper
    status = func(*args, **kwargs)
  File "/export/home/hysdsops/mozart/ops/opera-pcm/data_subscriber/daac_data_subscriber.py", line 37, in main
    run(sys.argv)
  File "/export/home/hysdsops/mozart/ops/opera-pcm/data_subscriber/daac_data_subscriber.py", line 70, in run
    results["download"] = run_download(args, token, es_conn, netloc, username, password, cmr, job_id)
  File "/export/home/hysdsops/mozart/ops/opera-pcm/data_subscriber/daac_data_subscriber.py", line 116, in run_download
    downloader.run_download(args, token, es_conn, netloc, username, password, cmr, job_id)
  File "/export/home/hysdsops/mozart/ops/opera-pcm/data_subscriber/asf_cslc_download.py", line 95, in run_download
    downloads = self.get_downloads(new_args, es_conn)
  File "/export/home/hysdsops/mozart/ops/opera-pcm/data_subscriber/asf_cslc_download.py", line 318, in get_downloads
    assert len(downloads) > 0, f"No downloads found for batch_id={current_batch_id}!"
AssertionError: No downloads found for batch_id=f24733_a2676!
```

[...]

Issue #1080, opened 2025-02-11
[Bug]: DISP-S1 Query Job in historical processing must validate all CSLC bursts


philipjyoon:
### Checked for duplicates

Yes - I've already checked

### Describe the bug

On rare occasion, less than 0.1% of the time, CMR may return incomplete amount of data. Seems to be a transient issue.
 
When running in historical mode the PCM query code currently filters out any bursts that don’t belong in burst patterns but does not validate that all patterns have sufficiently been retrieved. It does this when we’re in forward mode for triggering rule. We had assumed that CMR query result would always be consistent and since we’d had done the burst pattern grouping ahead of time in the database file, we wouldn't need to further validate. This is a bad assumption and could bite us if CMR has a large inconsistency issue in the future.
 
The current behavior is that the query and download jobs will succeed but SCIFLO will fail because the coverage is consistent. In fact, the query job should fail in historical processing mode.

The current work-around is to simply resubmit the query job and verify that all CSLC bursts are returned from CMR. This can be inspected by comparing the query job log output to `disp_s1_burst_db_tool.py`

### What did you expect?

n/t

### Reproducible steps

```bash
1.
2.
3.
...
```

### Environment

```bash
- Version of this software [e.g. vX.Y.Z]
- Operating System: [e.g. MacOSX with Docker Desktop vX.Y]
...
```

PR #1107, opened 2025-03-11
DISP-S1 historical processing query and download job final improvements

philipjyoon:
## Purpose
- Clear, easy-to-understand sentences outlining the purpose of the PR
## Proposed Changes
- Fixes for #1080
- Fixes for #1102
- Increased the machine sizes for CSLC query and download jobs because often the Sentinel1 process would take up an entire CPU core and not leave much for the actual functionality
- Tuned ASG for query and download jobs to make sure that query and download jobs aren't the bottleneck in processing when processing Priority 4 frames.

## Issues
- https://github.com/nasa/opera-sds-pcm/issues/1080
- https://github.com/nasa/opera-sds-pcm/issues/1102

## Testing
- Ran through the entire DISP-S1 Priority 1 Frames running query and download job and verified that all were successful.
- Ran the first set of processing for Priority 4 Frames which is the largest with ~800 frames. 
- Noticed that dev-sized GRQ ES couldn't keep up with the query load. Verified that production-sized GRQ is able to handle with 4x margin to spare.
- Wrote custom tool and performed validation for all Priority 1 Frames: https://github.com/nasa/opera-sds-pcm/blob/1106_disp_s1_hist_job_submission_validation/tools/testing/validate_cslc_downloads.py

### DIFF
--- a/cluster_provisioning/modules/common/variables.tf
+++ b/cluster_provisioning/modules/common/variables.tf
@@ -502,7 +502,7 @@ variable "queues" {
     }
     "opera-job_worker-cslc_data_query" = {
       "name"              = "opera-job_worker-cslc_data_query"
-      "instance_type"     = ["t3a.medium", "t3.medium", "t2.medium", "c6i.large", "t3a.large", "m6a.large", "c6a.large", "c5a.large", "r7i.large", "c7i.large"]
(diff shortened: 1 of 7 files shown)


collinss-jpl (approved):
A few suggestions, but looks good otherwise

PR #1134, opened 2025-04-15
DIST-S1 Switch from fixed inputs to triggered inputs

RKuttruff:
## Purpose
- Switch away from running DIST-S1 on demand with fixed inputs to jobs set up with RTC queries
## Proposed Changes
- [CHANGE] DIST-S1 chimera config populated by inputs from RTC query job
- [REMOVE] Canned DIST-S1 input function

### DIFF
--- a/opera_chimera/precondition_functions.py
+++ b/opera_chimera/precondition_functions.py
@@ -707,77 +707,61 @@ class OperaPreConditionFunctions(PreConditionFunctions):
 
         return rc_params
 
-    def get_dist_s1_sample_inputs(self):
+    def get_dist_s1_mgrs_tile(self):
         """
-        Temporary function to stage the "golden" inputs for use with the DIST-S1
-        PGE.
-        TODO: this function will eventually be phased out as functions to
-              acquire the appropriate input files are implemented with future
-              releases
+        Assigns the MGRS tile ID for DIST-S1 jobs
         """
         logger.info(f"Evaluating precondition {inspect.currentframe().f_code.co_name}")
 
-        # get the working directory
-        working_dir = get_working_dir()
+        metadata: Dict[str, str] = self._context["product_metadata"]["metadata"]
+
+        rc_params = {
+            'mgrs_tile_id': metadata['mgrs_tile_id'],
+        }
+
+        return rc_params
+
+    def get_dist_s1_rtc_s3_paths(self):
+        """
+        Gets the list of input RTCs to be processed by the DIST-S1 job and organizes them into
+        the pre-/post-copol/-crosspol lists
+        """
+        logger.info(f"Evaluating precondition {inspect.currentframe().f_code.co_name}")
+
+        metadata: Dict[str, str] = self._context["product_metadata"]["metadata"]
+
+        dataset_type = self._context["dataset_type"]
+
+        product_paths = metadata["product_paths"][dataset_type]
 
         rtc_pattern = re.compile(r'(?P<id>(?P<project>OPERA)_(?P<level>L2)_(?P<product_type>RTC)-(?P<source>S1)_'
-                                 r'(?P<burst_id>\w{4}-\w{6}-\w{3})_(?P<acquisition_ts>(?P<acq_year>\d{4})(?P<acq_month>'
-                                 r'\d{2})(?P<acq_day>\d{2})T(?P<acq_hour>\d{2})(?P<acq_minute>\d{2})(?P<acq_second>'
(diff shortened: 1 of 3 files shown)


data_subscriber/asf_rtc_for_dist_download.py
```diff
@@ -51,6 +51,8 @@ def run_download(self, args, token, es_conn, netloc, username, password, cmr,
 
         # WARNING: https download does not work
         # TODO: Should we support this at all?
+        #  Not sure how to split these paths into baseline & current lists - maybe we should until
```
collinss-jpl, line 54: Is this still relevant? Seems like the split is taken care of already below.

opera_chimera/precondition_functions.py
```diff
@@ -795,10 +779,106 @@ def sort_fn(path):
...
+
+        return rc_params
+
+    def get_dist_s1_lookbacks(self):
+        """
+        Get number of lookbacks for DIST-S1 job
+        """
+        logger.info(f"Evaluating precondition {inspect.currentframe().f_code.co_name}")
+
+        # This is currently hardcoded, should we move it to settings?
```
collinss-jpl, line 855: I think it's a safe bet that we'll want this exposed in settings.yaml

opera_chimera/precondition_functions.py
```diff
@@ -707,77 +707,61 @@ def get_dswx_ni_sample_inputs(self):
...
+        dataset_type = self._context["dataset_type"]
 
-        rtc_dates = sorted(os.listdir(rtc_data_dir))
+        product_paths: Dict[str, List[str]] = metadata["product_paths"][dataset_type]
+
+        rtc_pattern = re.compile(r'(?P<id>(?P<project>OPERA)_(?P<level>L2)_(?P<product_type>RTC)-(?P<source>S1)_'
+                                 r'(?P<burst_id>\w{4}-\w{6}-\w{3})_(?P<acquisition_ts>(?P<acquisition_date>'
+                                 r'\d{8})T\d{6}Z)_(?P<creation_ts>\d{8}T\d{6}Z)_(?P<sensor>S1A|S1B)_(?P<spacing>30)_'
+                                 r'(?P<product_version>v\d+[.]\d+))(_(?P<pol>VV|VH|HH|HV|VV\+VH|HH\+HV)|_BROWSE|_mask)?'
+                                 r'[.](?P<ext>tif|tiff|h5|png|iso\.xml)$')
```
collinss-jpl, line 743: Since all you really care about here is extracting the polarization flag, maybe its worth simplifying this regex to one that only looks to capture that part of the filename, something like:

`rtc_pattern = re.compile(r'.*_(?P<pol>VV|VH|HH|HV|VV\+VH|HH\+HV)_.*')`

opera_chimera/precondition_functions.py
```diff
@@ -795,10 +779,106 @@ def sort_fn(path):
...
+
+        dataset_type = self._context["dataset_type"]
+
+        product_paths: Dict[str, List[str]] = metadata["product_paths"][dataset_type]
+
+        rtc_pattern = re.compile(r'(?P<id>(?P<project>OPERA)_(?P<level>L2)_(?P<product_type>RTC)-(?P<source>S1)_'
+                                 r'(?P<burst_id>\w{4}-\w{6}-\w{3})_(?P<acquisition_ts>(?P<acquisition_date>'
+                                 r'\d{8})T\d{6}Z)_(?P<creation_ts>\d{8}T\d{6}Z)_(?P<sensor>S1A|S1B)_(?P<spacing>30)_'
+                                 r'(?P<product_version>v\d+[.]\d+))(_(?P<pol>VV|VH|HH|HV|VV\+VH|HH\+HV)|_BROWSE|_mask)?'
+                                 r'[.](?P<ext>tif|tiff|h5|png|iso\.xml)$')
```
collinss-jpl, line 872: Similar to above comment, can we reduce this regex to just an extraction of `acquisition_date`?

data_subscriber/asf_rtc_for_dist_download.py
RKuttruff: The comment is in reference to the case of `args.transfer_protocol == "https"`, that seems to be building its own set of s3 paths instead of the well split collections we get from CMR

opera_chimera/precondition_functions.py
RKuttruff: I did this but just stripped out the unused named groups

RKuttruff:
@hhlee445 Before you merge, I'm trying out some simplified regexes per Scott's feedback. I already found something I missed...

PR #1163, opened 2025-05-13
DISP-S1 PGE v3.0.6 Integration

### LINKED ISSUE #1156: [Task]: Integrate DIST-S1 6.0.0-er.3.0 PGE
### Checked for similar tasks

Yes - I've already checked

### Describe the task

Integrate the latest DIST-S1 PGE, exposing the multiprocessing parameters in settings.yaml

collinss-jpl:
## Purpose
- This branch integrates the DISP-S1 PGE v3.0.6 with OPERA PCM. This version continues to wrap the previous SAS delivery, but incorporates the DISP frame number into the output Compressed-CSLC file names to avoid potential collisions when adjacent frames share an input CSLC burst.

## Issues
- Resolves #1156 

## Testing
- Branch has been tested on a dev cluster by running the following:
  - DISP-S1 PGE workflow using simulation mode
  - DISP-S1 PGE workflow using real mode
  - DISP-S1 PGE smoke test
- Changes to the CCSLC output file name has been visually inspected in S3 and within the ES index for CCSLC products:

In S3
Screenshot 2025-05-13 at 12 27 12 PM

In ES
Screenshot 2025-05-13 at 12 35 29 PM

### DIFF
--- a/cluster_provisioning/dev-e2e/variables.tf
+++ b/cluster_provisioning/dev-e2e/variables.tf
@@ -365,9 +365,9 @@ variable "pge_releases" {
   default = {
     "dswx_hls" = "1.0.3"
     "cslc_s1"  = "2.1.1"
-    "rtc_s1"   = "2.1.1" 
+    "rtc_s1"   = "2.1.1"
     "dswx_s1"  = "3.0.2"
-    "disp_s1"  = "3.0.5"
+    "disp_s1"  = "3.0.6"
     "dswx_ni"  = "4.0.0-er.3.0"
     "dist_s1"  = "6.0.0-er.2.0"
     "tropo"    = "3.0.0-er.1.0-tropo"
@@ -494,7 +494,7 @@ variable "amis" {
 }
 
 variable "cnm_r_sqs_arn" {
-} 
+}
 
 variable "asf_cnm_s_id_dev" {
 }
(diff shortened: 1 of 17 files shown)


philipjyoon:
@collinss-jpl Everything looks good except that the `frame_id` field in compressed cslc ES catalog is now being populated with a string form (with F with zero-fill) This field was already being populated and being used by the PCM code as you can see in the screenshot. All current Compressed CSLC search and reasoning code expects this field to be an integer.

I couldn't figure out where in your changes is creating that field. Is it possible to not create that field at the PGE-level since PCM is already filling that field in and using it?

image

collinss-jpl:
> @collinss-jpl Everything looks good except that the `frame_id` field in compressed cslc ES catalog is now being populated with a string form (with F with zero-fill) This field was already being populated and being used by the PCM code as you can see in the screenshot. All current Compressed CSLC search and reasoning code expects this field to be an integer.
> 
> I couldn't figure out where in your changes is creating that field. Is it possible to not create that field at the PGE-level since PCM is already filling that field in and using it?
> image

Ah, this is happening because of the changes to the regular expression for CCSLC products that now calls out `frame_id` as one of the captured portions. These captured portions all wind up in the ES index under the same name, so the pre-existing `frame_id` field is getting clobbered. I will change the name of the variable in the regex to avoid this.

philipjyoon (approved):
Looks good

hhlee445 (approved):

PR #1210, opened 2025-07-16
Add setting for query & download of S1 SLCs from ESA Copernicus Dataspace instead of CMR/ASFDAAC

RKuttruff:
## Purpose
Due to recent issues with ASF, the need has become apparent to support an alternate source for our SLC inputs.

## Proposed Changes
- [ADD] Query and download options to source data from Dataspace
- [CHANGE] Rename some data subscriber base classes

## Tests
- [x] Temporal SLC query (ESA option set, Sim mode)
- [x] Native ID SLC query (ESA option set, Real mode)
  - PGEs executed with uncompressed ESA inputs:
    - [x] CSLC-S1
    - [x] CSLC-S1-STATIC
    - [x] RTC-S1
    - [x] RTC-S1-STATIC
- [x] All other product queries (ESA option set, Sim mode)
- [x] All product queries (ESA option unset, sim mode)
- [x] S1A/C Query timer 24-hour keep up
  - Insufficient capacity to keep up with 1 ESA login - will need further work
- [x] Real mode 3.2.0-rc1 test cases with ESA option
- [x] Real mode 3.2.0-rc1 test cases with CMR option


collinss-jpl (approved):
Changes look good to me! My only (optional) suggestion would be to include docstrings for the methods/functions in the brand new modules (`dataspace_download.py` and `esa_dataspace.py`). I realize most of the data subscriber still lacks docstrings, but no time like the present to start adding some, at least for new code.

hhlee445:
Missing `from data_subscriber.cmr import Provider` in slc_query.py

RKuttruff:
@hhlee445 Oops, I forgot to select that line for commit, should be fixed now

PR #1332, opened 2025-12-18
Install Duplicates check + DSWx-HLS accountability as metrics cron job

RKuttruff:
## Purpose
- Install duplicates + DSWx-HLS accountability scripts from nasa/opera-sds-ops#42 to metrics as a cron job running daily at midnight UTC

### DIFF
--- a/cluster_provisioning/dev-e2e/main.tf
+++ b/cluster_provisioning/dev-e2e/main.tf
@@ -107,6 +107,7 @@ module "common" {
   asf_cnm_s_id_test                       = var.asf_cnm_s_id_test
   asf_cnm_s_id_prod                       = var.asf_cnm_s_id_prod
   es_cluster_mode                         = var.es_cluster_mode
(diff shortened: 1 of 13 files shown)


RKuttruff:
@hhlee445 
How can we control whether this runs on a deployed cluster?
I'd like to avoid just commenting out the crontab entry by default, requiring a user to manually ssh into metrics to enable it.
Could we instead move it out of `conf/sds/files/metrics/cron/hysdsops` into a remote exec in terraform that only runs when a variable is set to true that appends the job to the crontab?

RKuttruff:
We might just be able to add a count parameter to [this](https://github.com/nasa/opera-sds-pcm/blob/f83ee83e108864223f4f4252a18da88f953a72ce/cluster_provisioning/modules/common/grq_factotum_metrics.tf#L210-L212) null resource that sets up the cron jobs on metrics

hhlee445 (approved):

PR #1394, opened 2026-04-09
Switch DSWx-NI SCIFLO inputs to triggered product metadata

RKuttruff:
## Purpose
- This PR replaces the fixed ADT test input for DSWx-NI jobs to the real triggered inputs.

## Issues
- [OPERA-2396](https://hysds-core.atlassian.net/browse/OPERA-2396)
## Testing
- [x] Cluster deploy
- [x] Real mode jobs
  - [x] Validate all files populate as expected
- [x] Regression test for DSWx-S1 (Repurposed some chimera functions for dual S1/NI use)


hhlee445 (approved):
Since we need to update the trigger logic for the frequencies and polarization. this PR can be merged

PR #1399, opened 2026-04-16
OPERA-2318: feat(DIST-S1): tile filter

chrisjrd:
tile filter params in hysds-io and job-spec
quiet (doubly) output of unzip
change job label for clarity
RTC CMR query by tile
print URL params to verify CMR request by visual inspection

NOTE: RTC metadata in CMR is not query-able by tile ID directly as there is no additional-property containing tile. This change uses a lookup table to map tiles to burst IDs for use as a CMR native-id pattern search.

## Purpose
Allow users to specify a list of tiles to limit query jobs. Especially for RTC for DIST-S1

## Proposed Changes
- [ADD] tile filter

## Issues
https://hysds-core.atlassian.net/browse/OPERA-2318

## Testing
tested on dev cluster


hhlee445 (approved):
LGTM

PR #1401, opened 2026-04-16
First-pass ASG Instance Type Updates

RKuttruff:
## Purpose
- For all ASG queues:
  - Removed instance types that differ from most other types in the queue (ie, if most types are c.2xlarge and there's a c.4xlarge or c.xlarge)
  - Removed instance types that have been more expensive in the cost analyses (cost per job >= 1 stdev above mean)
- For all queues with < 4 instance types, that were modified above and all SCIFLO queues (except for DSWxNI, DIST-S1, DISP-*, and Product Update) + some others:
  - Massively expanded instance type lists with all types similar to those already present (ie, family & size + intel/AMD if required)

## Issues
- First pass changes for [OPERA-2376](https://hysds-core.atlassian.net/browse/OPERA-2376), will re-evaluate against cost metrics after this is deployed in OPS


hhlee445 (approved):
LGTM

PR #1414, opened 2026-05-04
DIST-S1 Types - Adjust for cost

RKuttruff:
Drop c5* and m5* as they are the most expensive per-job

## Issues
- Links to relevant issues
- Example: issue-XYZ
## Testing
- [OPERA-2376](https://hysds-core.atlassian.net/browse/OPERA-2376). Adjustments made for DIST-S1 based on metrics data from INT test runs


hhlee445 (approved):
LGTM

PR #1415, opened 2026-05-04
DSWx-NI StateConfig-based Triggering Logic

RKuttruff:
## Purpose
Switch DSWx-NI triggering logic to use state configs

Related PR: nasa/opera-sds-lambdas#57

## Issues
- [OPERA-2380](https://hysds-core.atlassian.net/browse/OPERA-2380)

## Deploying for testing

In the .tfvars file, set the value of `lambda_package_release` to `OPERA-2380`

Further, due to how the GCOV beta input data was delivered, I recommend modifying the code of the deployed catalog ingest lambda (`opera-{venue}-{counter}-gcov-catalog-ingest-timer`) as follows, setting the values of `TEST_DT_BUCKET` and `TEST_DT_KEY` to the S3 location to store the temporary test file: 

<details>

<summary> Lambda code </summary>

```python
import argparse
import json
import logging
import os
from datetime import datetime, timezone, timedelta

import dateutil.parser
import requests
from dateutil.relativedelta import relativedelta


logger = logging.getLogger()
logging.basicConfig(
    level=logging.INFO,
    format='%(asctime)s [%(levelname)s] [%(name)s::%(lineno)d] %(message)s'
)
logger.setLevel(logging.INFO)

DATETIME_FORMAT = "%Y-%m-%dT%H:%M:%SZ"
JOB_NAME_DATETIME_FORMAT = "%Y%m%dT%H%M%S"

TEST_START_DT = datetime(2026, 2, 28, 7)
TEST_MAX_DT = datetime(2026, 2, 28, 8)
TEST_DT_BUCKET = '<dev bucket>'
TEST_DT_KEY = '<test key>'
TEST_DELTA = timedelta(minutes=2)
TEST_DT_FMT = '%Y-%m-%dT%H:%M:%S'

import boto3
s3 = boto3.client('s3')


def pull_start_dt():
    try:
        dt_str = s3.get_object(
            Bucket=TEST_DT_BUCKET,
            Key=TEST_DT_KEY
        )['Body'].read().decode('utf-8')
        dt = datetime.strptime(dt_str, TEST_DT_FMT)
    except Exception as e:
        print(f'Could not pull start time: {e}. Using default start time.')
        dt = TEST_START_DT
    
    if dt >= TEST_MAX_DT:
        raise Exception('Reached end of test range')

    print(f'Using start time: {dt}')
    return dt


def push_start_dt(dt):
    print(f'Pushing new start time {dt}')
    dt_str = dt.strftime(TEST_DT_FMT)
    s3.put_object(
        Bucket=TEST_DT_BUCKET,
        Key=TEST_DT_KEY,
        Body=dt_str.encode('utf-8')
    )


def submit_job(
        mozart_host,
        job_type,
        job_release,
        job_queue,
        priority,
        tags,
        query_end_dt: datetime,
        # use_temporal,
        minutes,
        revision_margin,
        dedup,
        **extra_params
):
    mozart_url = f'https://{mozart_host}/mozart/api/v0.1/job/submit'

    job_params: dict = {}

    query_end_dt = query_end_dt - relativedelta(minutes=revision_margin)
    # query_start_dt = query_end_dt - relativedelta(minutes=minutes)
    query_start_dt = query_end_dt - TEST_DELTA # TEMP OVERRIDE DELTA

    # *_PARAM = param name in HySDS IO, *_PARAM_PREFIX useful if param values are CLI parameters (eg. --start-date=)
    job_params[os.getenv('START_PARAM', 'start_date')] = f'{os.getenv('START_PARAM_PREFIX', '')}{query_start_dt.strftime(DATETIME_FORMAT)}'
    job_params[os.getenv('END_PARAM', 'end_date')] = f'{os.getenv('END_PARAM_PREFIX', '')}{query_end_dt.strftime(DATETIME_FORMAT)}'

    job_params.update(extra_params)

    payload = {
        'queue': job_queue,
        "priority": priority,
        'tags': json.dumps([tags]),
        'type': f'job-{job_type}:{job_release}',
        'params': json.dumps(job_params),
        'name': f'catalog-ingest-timer-{job_type}-{datetime.now(timezone.utc).replace(tzinfo=None).strftime(JOB_NAME_DATETIME_FORMAT)}_{minutes}',
        # 'enable_dedup': dedup
    }

    logger.info(f'Mozart payload: {json.dumps(payload)}')
    logger.info(f'Submitting job to {mozart_url}')

    resp = requests.post(
        mozart_url,
        data=payload,
        params=dict(enable_dedup=dedup),
        # headers={'Content-Type': 'application/json'},
        verify=False
    )

    logger.info(f"Request code: {resp.status_code}")
    logger.info(f"Request text: {resp.text}")
    resp.raise_for_status()

    result = resp.json()

    if "result" in result.keys() and "success" in result.keys():
        if result["success"] is True:
            job_id = result["result"]
            logger.info(f"submitted job: {job_type} job_id: {job_id}")
            return job_id
        else:
            logger.error(f"job not submitted successfully: {result}")
            raise Exception(f"job not submitted successfully: {result}")
    else:
        raise Exception(f"job not submitted successfully: {result}")


def lambda_handler(event, context):
    logger.info(f"Got event: {json.dumps(event)}")

    mozart_host = os.environ['MOZART_HOST']
    job_release = os.environ['JOB_RELEASE']

    job_type = event.pop('job_type')
    job_queue = event.pop('job_queue')
    priority = event.pop('priority')
    tags = event.pop('tags')
    minutes = event.pop('minutes', 60)
    revision_margin = event.pop('revision_margin', 0)
    extra_params = event.pop('extra_params', {})
    dedup = event.pop('enable_dedup', False)

    event_time = event.pop('time')

    test_start_dt = pull_start_dt()
    test_end_dt = test_start_dt + TEST_DELTA

    submit_job(
        mozart_host,
        job_type,
        job_release,
        job_queue,
        priority,
        tags,
        # dateutil.parser.isoparse(event_time),
        test_end_dt,  # OVERRIDE START TIME
        minutes,
        revision_margin,
        dedup,
        **extra_params
    )

    push_start_dt(test_end_dt)


if __name__ == '__main__':
    parser = argparse.ArgumentParser()

    parser.add_argument('mozart_host', help='Mozart host (IP or hostname with optional port)')
    parser.add_argument('job_type', help='Type of job to submit')
    parser.add_argument('job_release', help='Job release')
    parser.add_argument('job_queue', help='Job queue')
    parser.add_argument('priority', type=int, help='Job priority')
    parser.add_argument('tags', help='Tags')
    parser.add_argument('minutes', type=int, help='minutes')
    parser.add_argument('revision_margin', type=int, help='revision_margin')
    parser.add_argument('--dedup', action='store_true', help='Enable job dedupe')

    args = parser.parse_args()

    submit_job(
        args.mozart_host,
        args.job_type,
        args.job_release,
        args.job_queue,
        args.priority,
        args.tags,
        datetime.now(timezone.utc),
        args.minutes,
        args.revision_margin,
        args.dedup,
    )
```
</details>

These changes override the query start and end times to go from `TEST_START_DT` to `TEST_MAX_DT`, stepping in increments of `TEST_DELTA` minutes. This is important to avoid triggering many thousands of jobs with a single query, and better approximates running the timer in forward mode against the live GCOV dataset.

## Testing
- [x] Cluster deploy
- [x] Test cases (catalog ingest -> evaluator -> sciflo)
  - [x] Full coverage
  - [x] Partial coverage - timeout
  - [x] Partial coverage - resubmit on new input
  - [x] Ingest timer
  - [x] Expiry timer


RKuttruff:
## Ran a test using the timer lambdas

Ran the following GCOV revision time query ranges:
- Main timer range
  - 2026-02-28T07:00:00Z - 2026-02-28T08:00:00Z (2 minute increments)
  - 4,161 `L2_GCOV_NI` docs created
- Secondary test (SC update after expiration)
  - 2026-01-20T11:55:49Z - 2026-01-20T11:56:20Z, then 2026-01-20T11:56:23Z - 2026-01-20T11:56:23Z
  - 1 `L2_GCOV_NI` for each

Created 4,690 State configs
- 2,878 marked as complete
  - 1,557 complete and not skipped
- 1,810 marked as expired (+2 from secondary test which were unmarked after the second query)
  - 857 expired and not skipped (+2 from secondary test which were unmarked after the second query)
- 2,274 marked as skipped (no valid granules)

Triggered 2,417 DSWx-NI SCIFLO jobs
- 1,557 full jobs
- 859 partial jobs
- 1,109 deduped jobs (suspect to be caused by the completed SC being published multiple times by parallel evaluator jobs)
- 232 jobs failed
  - +2 additional transient GDAL failures that were retried successfully
  - 8 failed without a linked triage (didn't check bucket) with `celery.utils.serialization.UnpickleableExceptionWrapper: WorkerExecutionError('SciFlo step L3_DSWx_NI_PGE failed: PGE/SAS subprocess failure\n')`
    - Retried (in hopes to resolve a proper error): All still failed but with usable error messages and linked triage data
  - Error breakdown:
    - 19 OOM issues
      - All in jobs with full coverage
      - Jobs were retried with 16 GiB of Verdi swap space added: all jobs completed
    - 211 Jobs produced no output GeoTIFFs, leading to a PGE postproc failure
      - 199 were in partially covered jobs
      - 12 were in fully covered jobs
    - 2 failures due to `"ValueError: cannot convert float NaN to integer` during GCOV mosaic
      - Both in partial jobs

After OOM reruns (2,203 in `job-completed`):
- 17,517 L2_DSWx_NI products published

hhlee445:
Getting this error from _**trigger-send_cnm_notify_msg_L3_DSWX_NI**_ job.  

```
Traceback (most recent call last):
  File "/root/verdi/ops/hysds-3.1.1/hysds/job_worker.py", line 1137, in run_job
    image_info = container_engine.ensure_image_loaded(
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/root/verdi/ops/hysds-3.1.1/hysds/containers/base.py", line 315, in ensure_image_loaded
    raise RuntimeError(
RuntimeError: Failed to download image s3://s3-us-west-2.amazonaws.com/opera-dev-cc-fwd-hyunlee/container-nasa_opera-sds-pcm:issue_2380_dswx-ni-evaluator.tar.gz:
[Errno 28] No space left on device
```

**_`timer-GCOV-catalog-ingest`_**

```
Traceback (most recent call last):
  File "/root/verdi/ops/hysds-3.1.1/hysds/job_worker.py", line 1137, in run_job
    image_info = container_engine.ensure_image_loaded(
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/root/verdi/ops/hysds-3.1.1/hysds/containers/base.py", line 315, in ensure_image_loaded
    raise RuntimeError(
RuntimeError: Failed to download image s3://s3-us-west-2.amazonaws.com/opera-dev-cc-fwd-hyunlee/container-nasa_opera-sds-pcm:issue_2380_dswx-ni-evaluator.tar.gz:
[Errno 28] No space left on device
```

RKuttruff:
@hhlee445 I was experimenting with adding swap space on the verdi workers. It looks like this runs out the disk space for smaller jobs, similar to what was happening with mozart. I've reverted the changes, so these should work if you redeploy or modify the existing launch templates with the change in bccbc70

data_subscriber/gcov/gcov_mgrs_evaluator.py
```diff
@@ -0,0 +1,469 @@
...
+
+        existing_document = backoff_wrapper(
+            self.es_conn.search_by_id,
+            id=sc_id,
+            index=c.MGRS_SET_STATE_CONFIG_ES_PATTERN,
+            ignore=[404]
+        )
+
+        if existing_document.get("found", False):
+            metadata = existing_document.get('_source', {}).get('metadata', {})
```
sjlewis-jpl, line 435: The return arguments repeats `IS_EXPIRED` for the first two return arguments.  When it's called, the return arguments are `complete, expired, skipped = self._get_state_config_state(sc_id)`.  I think the first return argument should be `IS_COMPLETE`?

data_subscriber/gcov/gcov_catalog_ingest.py
```diff
@@ -0,0 +1,250 @@
...
+                try:
+                    result = es_conn.es.search(
+                        index="grq_*_l2_gcov_ni-*",
+                        body={"query": {"term": {"_id": granule_ur}}, "size": 0},
+                    )
+                    if result["hits"]["total"]["value"] > 0:
+                        skipped += 1
+                        logger.info(f'Skipping granule {granule_ur} as it has been ingested already')
+                        continue
+                except Exception as e:
```
sjlewis-jpl, line 152: It looks like the Exception handles the case where the record is not found, and in that case creates the new record of the input.  But if there's a connection error with the database, that might cause a duplicate record to be generated.  Is there a more specific error we could catch that means everything is working and the record is not found?  Or perhaps consider catching errors like ConnectionError, Timeout, TransportError, etc., and perhaps raise an actual error, or retry, or something sensible.

data_subscriber/gcov/gcov_mgrs_evaluator.py
```diff
@@ -0,0 +1,469 @@
...
+                else:
+                    logger.warning(f'Unexpected track frame: {track_frame} for query {body}')
+
+        found_track_frames = list(found_track_frames)
+        found_track_frames.sort()
+
+        excluded_track_frames = list(excluded_track_frames)
+        excluded_track_frames.sort()
+
+        return found_track_frames, excluded_track_frames, product_paths, min(start_times), max(end_times)
```
sjlewis-jpl, line 287: Is there a chance for the ES query to return no GCOV results (e.g., all products filtered out by polarization/bandwidth checks, or aren't in the expected_track_frames)?  If that situation, the `start_times` and `end_times` lists will be empty, causing ValueError on min/max.  That would crash this function... That seems like it will crash the `evaluate()` without creating a state config, resulting in missing DSWx-NI products that should be tracked as incomplete.

My suggestion:  add a guard before return: `if not start_times: raise ValueError(f\"No valid GCOV found for {expected_track_frames}, cycle {cycle_number}\")`, or return sentinel values that _create_sc can handle. Explicit 'no products found' handling is better than silent failure.

data_subscriber/gcov/gcov_mgrs_evaluator.py
```diff
@@ -0,0 +1,469 @@
...
+        start_times = []
+        end_times = []
+
+        if results:
+            for hit in results:
+                source = hit.get("_source", {})
+                meta = source.get("metadata", {})
+                track_frame = meta.get("track_frame")
+                if track_frame and track_frame in expected_track_frames:
+                    start_times.append(source['starttime'])
```
sjlewis-jpl, line 266: Should the state config span all expected products?  Or just the valid ones?  If the latter, an inline comment explaining that would be useful.  Currently, it looks like the start times and end times are only collected from the valid products.  If the state config should span all expected products, then the logic for appending start and end times should be moved outside the checks.

data_subscriber/gcov/gcov_mgrs_evaluator.py
```diff
@@ -0,0 +1,469 @@
...
+        try:
+            logger.info(f'Attempting refresh on {index_pattern}')
+            self.es_conn.es.indices.refresh(
+                index=index_pattern,
+                ignore_unavailable=True,
+                allow_no_indices=True,
+                expand_wildcards="open",
+            )
+        except Exception as e:
+            logger.warning(f'Failed index refresh: {e}')
```
sjlewis-jpl, line 60: This looks like it will hide failures that the pipeline should surface, but I'm uncertain of this functions job.  If that's the intended behavior, I suggest adding an inline comment to explain why we're not reraising this error.

If just giving a warning is the right move here, I suggest being more verbose in the log message.  During testing, I often see minimal warnings and have no idea whether they are significant.  Write the message for Lala and the operators to understand.

data_subscriber/gcov/mgrs_track_collections_db.py
```diff
@@ -290,3 +315,34 @@ def get_bounding_box_for_mgrs_set_id(self, mgrs_set_id):
         )
 
         return [xmin, ymin, xmax, ymax]
+
+    @cache
+    def get_geojson_for_mgrs_set_id(self, mgrs_set_id):
```
sjlewis-jpl, line 320: Add a one-line docstring describing the intent and I/O.

sjlewis-jpl (commented):
I'm leaving this as a comment review, though I think the inline comments should be addressed.

The biggest issue I see, though, is a lack of any test files with this PR.  This is major component of the DSWx-NI pipeline.  Most of the issues I point out inline could be caught by good unit tests.  I do see some test data added as a binary file, so perhaps there is some other testing that isn't part of these changes?

Docstrings for the functions would help a TON for understanding what each function does, especially for inputs and outputs.  One of the issues flagged had an issue with the outputs, which required tracing it backwards (not easily done in review mode).  This will be pretty important logic in the near future, so the documentation could really help with debugging or modifying once we get into system testing and products validation.

For the public functions in the class, I would say docstrings are critical.

There a few comments about exceptions and log messages too.  Any log message above debug should be targeted at our operators.  If Lala and the Ops crew wouldn't understand the message, there needs to be more written into the log.  Debug messages belong to the developers.  Anything more is actually a message for Operators.

data_subscriber/gcov/gcov_mgrs_evaluator.py
RKuttruff: This should be an extremely rare failure - I've never seen a refresh call fail.

By default, indices are auto-refreshed every second, so the consequence of a failed refresh would be docs less than 1 second old may not be returned in queries. There are 2 index patterns that are refreshed by this method: the GRQ state config index and the GRQ GCOV index. The state config index is queried to check for completeness of the current state config, and double-checked to ensure a parallel evaluator didn't mark the sc as complete during the runtime of the current evaluator, which would submit a duplicate job (which gets deduped by HySDS anyway). A failed refresh for this would make this check less reliable (and even with the check it's pretty unreliable, hence why I intend to do a little refactoring after this PR is merged which would make this refresh even less critical). For the GCOV index, this checks for current coverage of the current state config. If the refresh fails, and documents are missed due to an ingest job actively publishing new GCOVs, then the sc will be marked as partially complete and no further action will be taken. The missed GCOVs will themselves trigger new evaluator jobs so they won't be missed if the refresh fails here.

In any case, I'll add exponential backoff to the call (which, if that fails out, there's probably much bigger problems than the failed refresh would cause) and expand on the warning message.

RKuttruff: So the tl;dr of this refresh call is it's trying to reduce the number of deduped SCIFLO jobs, but may be superseded by an upcoming refactor.

data_subscriber/gcov/gcov_mgrs_evaluator.py
RKuttruff: > Should the state config span all expected products?

This is essentially what is happening here, the pol/mode validity checks happen a few lines later. This check doubly ensures that only GCOVs that map to to MGRS tile set being evaluated are considered, which is primarily done by the query above

data_subscriber/gcov/gcov_mgrs_evaluator.py
RKuttruff: This shouldn't happen. As stated [here](https://github.com/nasa/opera-sds-pcm/pull/1415#discussion_r3540282586), GCOVs skipped for pol/mode still contribute to start/end time (though I wonder if they should, may be a question worth asking). There shouldn't be a situation where this query returns nothing, as this job is triggered by such GCOVs existing. I suppose this could happen if someone manually crafted an empty state config dataset, ingested it, and then triggered the evaluator job on it, but it would just fail anyway.

data_subscriber/gcov/gcov_mgrs_evaluator.py
RKuttruff: Good catch, thanks! I've fixed this in the upcoming PR

data_subscriber/gcov/gcov_catalog_ingest.py
RKuttruff: I can put this in a backoff wrapper. I think the only consequence of this check failing and publishing a duplicate (or rather overwriting an existing record with the same data) is that it would trigger an extra evaluator job that would effectively be a no-op (skipping if the set is complete or finding no new granules and not changing anything)

sjlewis-jpl: Ok, I hear that.  I think an inline comment describing the minor consequence would be useful, rather than relying on people inferring that lack of consequence in the future. (I guess doing that in the next PR, since this one already got merged)

data_subscriber/gcov/gcov_mgrs_evaluator.py
sjlewis-jpl: You're right, my apologies for the false alarm.

data_subscriber/gcov/gcov_catalog_ingest.py
RKuttruff: How's this?

```python
logger.warning(f"GRQ check failed for {granule_ur}: {e}. Proceeding with creation. This may trigger "
               f"an extra state config evaluation, but this should have no negative impacts.")
```

sjlewis-jpl: Much better, thank you!

PR #1420, opened 2026-05-12
OPERA-0000: rtc dedupe

chrisjrd:
## Purpose
reduce chance of duplicate inputs being used in DIST-S1
## Proposed Changes
- [CHANGE] duplicate detection

## Issues
TBD
## Testing
tested locally

### DIFF
--- a/data_subscriber/rtc_for_dist/rtc_for_dist_query.py
+++ b/data_subscriber/rtc_for_dist/rtc_for_dist_query.py
@@ -72,12 +72,12 @@ class RtcForDistCmrQuery(BaseQuery):
         '''
         granules_dict = {}
         for granule in granules:
-            key = (granule["burst_id"], granule["acquisition_ts"][:-5])  # within the same hour of day (format 'YYYYMMDDTHH') to mitigate duplicates that are within 1 second.
+            key = (granule["burst_id"], granule["acquisition_ts"].strftime("%Y%m%dT%H"))[:-5]
             if key not in granules_dict:
                 granules_dict[key] = granule
             else:
                 self.logger.debug(f"Found duplicate granules {granule['granule_id']}, {granules_dict[key]['granule_id']} with the same burst_id and acquisition_ts. Keeping only the latest production one.")
-                if granule["acquisition_ts"][:-5] > granules_dict[key]["acquisition_ts"][:-5]:
+                if granule["acquisition_ts"].strftime("%Y%m%dT%H%M%S")[:-5] > granules_dict[key]["acquisition_ts"].strftime("%Y%m%dT%H%M%S")[:-5]:
                     granules_dict[key] = granule
         return list(granules_dict.values())
 
--- a/data_subscriber/rtc_for_dist/rtc_for_dist_query.py
+++ b/data_subscriber/rtc_for_dist/rtc_for_dist_query.py
@@ -66,15 +66,18 @@ class RtcForDistCmrQuery(BaseQuery):
                 raise AssertionError("--product-id-time must be provided in DIST-S1 reprocessing mode.")
 
     def unique_latest_granules(self, granules):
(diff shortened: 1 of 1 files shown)


hhlee445 (approved):

hhlee445:
Verified with

```
python ~/mozart/ops/opera-pcm/data_subscriber/daac_data_subscriber.py query --collection-shortname=OPERA_L2_RTC-S1_V1 --product=DIST_S1 --endpoint=OPS \
--job-queue=opera-job_worker-rtc_for_dist_data_download --chunk-size=1  --use-temporal --transfer-protocol=auto --processing-mode=reprocessing --product-id-time=32VLP_5,20260107T055557Z


python ~/mozart/ops/opera-pcm/data_subscriber/daac_data_subscriber.py query --collection-shortname=OPERA_L2_RTC-S1_V1 --product=DIST_S1 --endpoint=OPS \
--job-queue=opera-job_worker-rtc_for_dist_data_download --chunk-size=1  --use-temporal --transfer-protocol=auto --processing-mode=reprocessing --product-id-time=32VKP_4,20260107T055557Z


python ~/mozart/ops/opera-pcm/data_subscriber/daac_data_subscriber.py query --collection-shortname=OPERA_L2_RTC-S1_V1 --product=DIST_S1 --endpoint=OPS \
--job-queue=opera-job_worker-rtc_for_dist_data_download --chunk-size=1  --use-temporal --transfer-protocol=auto --processing-mode=reprocessing --product-id-time=31VEJ_4,20260107T055557Z
```

data_subscriber/rtc_for_dist/rtc_for_dist_query.py
```diff
@@ -66,15 +66,18 @@ def validate_args(self):
...
     def unique_latest_granules(self, granules):
-        ''' Remove duplicate granules defined by having the same burst_id and acquisition_ts, keep just the latest one'''
+        ''' Remove duplicate granules defined by having the same burst_id and acquisition_ts, keep just the latest one
+
+        On rare occassion, duplicate granules may share acquisition ts within 1 second of each other.
+        '''
         granules_dict = {}
         for granule in granules:
-            key = (granule["burst_id"], granule["acquisition_ts"])
+            key = (granule["burst_id"], granule["acquisition_ts"].strftime("%Y%m%dT%H"))[:-5]
```
sjlewis-jpl, line 75: I'm guessing `granule["acquisition_ts"]` is a datetime object?

So, if I do this:  `datetime.fromisoformat('20260628T104912Z').strftime("%Y%m%dT%H")`, the result is `'20260628T10'`
Then adding the slice:  `datetime.fromisoformat('20260628T104912Z').strftime("%Y%m%dT%H")[:-5]` gives the result `'202606'`.
That looks like we're just looking at the year and month of the granules.

Perhaps you meant to call `strftime("%Y%m%dT%H%M%S")`?  Though that would yield '20260628T1', which still doesn't get to +/- 1 second.  Is there something I'm missing on this line?

chrisjrd: you're looking at the wrong line. This line is for generating the key of the dictionary, which essential groups by day+burst.

You're comment is more applicable to line 80. Note that on line 80 we use the full HMS format you expect.

Here are my remarks on line 80:

It is not meant to get down to the +/- 1 second, to keep the logic simple. The note in the docstring is for context: it is not meant to describe what the code is doing.

So the slice is meant to compare acquisition dates down to the first hour digit.
We slice out the seconds because of the potential wrap on the tenths place of the seconds fragment. (:59 vs 1:00; :09 vs :10, etc.).
We slice out the minutes because of the potential wrap on the minutes fragment.
We slide out at least the last digit of the hours fragment for similar reasons.

So acquisitions within the same AM/PM or 12-hour section are considered duplicitous. It's not an accurate (100%), but we deemed practical at the time (99.999...% x w/e real duplicate acquisitions are).

PR #1421, opened 2026-05-12
DIST-S1 delete highly-replicated fields from dataset met

RKuttruff:
## Purpose
Similar to DISP-S1, remove the following from the dataset metadata:
- `runconfig.localize` and `runconfig.input_file_group.input_file_paths` as they're equivalent to lineage
- For each entry in `files`, remove `runconfig` and `lineage` as copies of these items exist in the top-level met
- In lineage, stripped out the common path prefix to a separate met entry

With DIST-S1, these removed fields can be very large due to hundreds of RTC inputs per DIST product. This unfortunately does not solve the issues with the mozart job status indices, as they contain many copies of the job context & product metadata (massive RTC lists). I believe this may be a HySDS-level issue.

## Issues
- Contributes to [OPERA-2482](https://hysds-core.atlassian.net/browse/OPERA-2482)
## Testing
- Verified reduction in GRQ document size

### DIFF
--- a/product2dataset/product2dataset.py
+++ b/product2dataset/product2dataset.py
@@ -347,12 +347,31 @@ def convert(
                 for l in dataset_met_json["lineage"]:
                     lineage_arr.append(l.split('/')[-1])
                 dataset_met_json["lineage"] = lineage_arr
+        elif pge_name == 'L3_DIST_S1':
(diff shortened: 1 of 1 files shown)


hhlee445 (approved):
LGTM

PR #1433, opened 2026-05-21
Use gp3 EBS volumes

RKuttruff:
Switch from gp2 -> gp3 EBS volumes. These are cheaper and have better performance

[OPERA-2500](https://hysds-core.atlassian.net/browse/OPERA-2500)


hhlee445 (approved):
LGTM

PR #1435, opened 2026-05-26
Ensure map staging fails with GDAL errors

RKuttruff:
## Purpose
Fully ensure against silent GDAL failures in the tiled ancillary map staging tools.

- Explicitly re-call `gdal.UseExceptions()` at the start of each download function
- Check each expected GDAL output file exists after each GDAL call, raising a `RuntimeError` if not

## Issues
- [OPERA-2499](https://hysds-core.atlassian.net/browse/OPERA-2499)
## Testing
- Tested with a DIST-S1 SCIFLO job with a custom UMD water mask input with a corrupted tile over the DIST tile being generated. An exception was properly raised.


hhlee445 (approved):
LGTM

PR #1436, opened 2026-05-27
OPERA-2498: Add delete_dataset.py for tile-driven DIST-S1 cleanup

pymonger:
## Summary

Adds a self-contained CLI for tile-driven cleanup of DIST-S1 datasets from GRQ (OpenSearch) and S3. Given a text file of MGRS tiles (one per line), it finds the matching DIST-S1 product docs and state-config docs, deletes their S3 prefixes, and removes the ES docs — replacing the manual, error-prone cleanup of invalid/duplicated/reprocessed tiles.

Resolves [OPERA-2498](https://hysds-core.atlassian.net/browse/OPERA-2498).

## Changes

**New tool**
- `tools/delete_dataset.py` — single-file utility (no new package deps beyond `boto3` / `opensearch-py` / `pyyaml`, all already in the PCM env):
  - **Input:** `--tile-file` (one MGRS tile per line; leading `T` and `#` comments tolerated; deduped).
  - **Selection:** matches `metadata.mgrs_tile_id.keyword` **or** `metadata.tile_id.keyword` (older state-configs only carry `tile_id`), against state-config index `grq_*dist_s1*state-config*` and product index `grq_*_l3_dist_s1*`.
  - **S3:** reads each product doc's `urls` + `browse_urls`, normalizes OPERA's `s3://<endpoint>:80/<bucket>/<key>` form, and deletes every object under the prefix (paginated `list_objects_v2` + batched `delete_objects`, parallelized across `--workers`). State-configs have no S3 and are ES-only.
  - **ES:** `delete_by_query` on `_id` in chunks of 1000, scoped per physical index.
  - **Safety:** dry-run by default; `--execute` to act; typed `DELETE` confirmation unless `--yes`; JSONL manifest of every resolved target written **before** any destructive call; `--max` hard cap (default 10000) with `--force` override; rolling `--log-file` audit trail; exit code `2` on any partial failure.
  - **Config:** reads `~/.sds/config` (`GRQ_ES_PROTOCOL` / `GRQ_ES_PVT_IP` / `GRQ_ES_PORT`), using `~/.netrc-os` when the protocol is `https`; overridable via `--es-url` / `--es-user` / `--es-password` / `--verify-certs`.

## Test plan

- [x] `--help` and dry-run smoke-tested on dev (gmanipon, https+netrc-os) and ops-fwd (http, no auth) — both auth paths connect.
- [x] URL normalization unit-checked for all four URL shapes (`s3://bucket/key`, `s3://endpoint:80/bucket/key`, `http(s)://bucket.s3-website…`, `http(s)://bucket.s3.region…`).
- [x] Single-tile execute on ops-pop1 (`01WCN`): pre-counts 4 state-config + 5 product → deleted 9 docs + 150 S3 objects across 10 prefixes in 1.2s; ES and S3 confirmed empty afterward, 0 errors.
- [x] Regression for older docs: 3 `_1_S1C_*` state-configs carrying only `metadata.tile_id` were initially missed, fix applied (both-field `should` query), re-validated — Tosca `query_string:"01WCN"` on the state-config index now returns 0.
- [x] Full 855-tile dry-run on ops-pop1 resolves 19,205 targets (10,038 state-config + 9,167 product, 18,334 S3 prefixes); hard cap correctly blocks the run without `--force`.
- [x] Manifest written before deletion and exit-code-2-on-partial-failure paths exercised.


hhlee445 (approved):
LGTM

hhlee445:
@pymonger should we rename this file as delete_dist_s1_dataset.py for now until we generalize the script for other dataset types?

pymonger:
Good call — done in 79147a9c (rename) + 28f7bfd7 (docstring/logger/manifest name sync). I also kept the file's selectors hard-coded to DIST-S1 only, so the narrower name matches what it actually does today. When we generalize for other dataset types we can decide between adding a `delete_dataset.py` umbrella or per-product scripts.

hhlee445 (approved):

PR #1437, opened 2026-05-27
OPERA-2508: single polarization skips in historical mode

chrisjrd:
## Purpose
address paused state-config processing for DIST-S1 historical runs.

## Proposed Changes
- [CHANGE] write a state-config marked skipped in the single polarization case.

## Issues
https://hysds-core.atlassian.net/browse/OPERA-2508
## Testing
tested on ops pop1 mozart


RKuttruff (approved):
LGTM

PR #1439, opened 2026-06-02
Add optional burst filtering for L2 products

RKuttruff:
## Purpose
This PR adds the option to only process specific bursts within a given S1 SLC input when generating RTC-S1, CSLC-S1 or the associated static layers products.

To use the filter, use the `--burst-ids` param with the DAAC data subscriber tool, giving one or more burst IDs to restrict to. Provided burst IDs are case insensitive and do not distinguish between `_` and `-` (in other words, both `t001_123456_iw1` and `T001-123456-IW1` are acceptable). No checks can be made to ensure the provided burst list is actually present within the SAFE file, and thus would proceed to the SAS to fail. This option is only available when triggering by native ID.

Example:
```bash
python3 ~/mozart/ops/opera-pcm/data_subscriber/daac_data_subscriber.py query -c SENTINEL-1A_SLC --job-queue=opera-job_worker-slc_data_download --chunk-size=1 --processing-mode=reprocessing -p ASF --native-id=S1A_IW_SLC__1SDV_20260420T171823_20260420T171850_064166_08137D_D6D2* --burst-ids T044-093062-IW3 T044-093061-IW3 T044-093061-IW2 T044-093063-IW1
```

## Issues
- [OPERA-2435](https://hysds-core.atlassian.net/browse/OPERA-2435)

## Testing
- [x] RTC-S1
- [x] CSLC-S1
- [x] RTC-S1-STATIC
- [x] CSLC-S1-STATIC

Results from running the above example command:

image

### DIFF
--- a/conf/RunConfig.yaml.L2_CSLC_S1.jinja2.tmpl
+++ b/conf/RunConfig.yaml.L2_CSLC_S1.jinja2.tmpl
@@ -71,9 +71,9 @@ RunConfig:
             {%- endif %}{# if runconfig.input_file_group[ type ] is iterable and runconfig.input_file_group[ type ] is not string #}
             {%- endif %}{# if runconfig.input_file_group[ type ] is not none #}
             {%- endfor %}{# for type in runconfig.input_file_group.keys() #}
-            {%- if runconfig.input_file_group.burst_list %}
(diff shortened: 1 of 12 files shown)


hhlee445 (approved):
LGTM

PR #1451, opened 2026-06-24
Fix CMR burst-audit off-by-one tracks at SLC boundaries (annotation-only derivation)

pymonger:
## Summary

Fix a longstanding bug in `cmr_audit_burst_coverage.py` (and the supporting `slc_annotation_extract.py`) that produced wrong burst-IDs for SLCs whose ASF SLC-BURST API response included bursts from adjacent acquisitions on a different relative-orbit track. Switches burst-ID derivation to use only the SLC's own annotation XMLs + `manifest.safe`, with a pure-Python port of the ESA formula (matches OPERA's PGE bit-perfect via `s1reader.S1BurstId.from_burst_params()`). Also includes a small fix making CMR 5xx errors retryable (instead of silently caching as "no products found").

## Background

The CMR burst-coverage audit tool reports per-burst missing OPERA RTC/CSLC products for a given date range, so we can request reprocessing of the responsible SLCs. The tool used the ASF SLC-BURST API to look up burst IDs per SLC; that API's time-range query intermittently returns bursts from adjacent SLC acquisitions. When those adjacent bursts were on a different relative-orbit track, our code anchored its annotation-XML derivation on a burst from the wrong SLC, producing off-by-one track numbers (e.g. T131 instead of T130). The audit then looked for the wrong burst-ID pattern in CMR and reported real products as missing while ignoring the actual products that did exist.

Concrete example caught by an OPS user: SLC `S1A_..._063027_07E8C5_7795` actually has 27 RTC products in CMR spanning tracks T130 / T131 (genuine ANX-crossing acquisition). Old code reported 7 of those bursts as "missing"; new code derives all 27 burst IDs identically to what is published in CMR.

### `tools/ops/cmr_audit/slc_annotation_extract.py`
- `extract_slc_metadata(zip_url, token)` — fetches annotation XMLs **and** `manifest.safe` via HTTP range requests (~50 KB extra; total still ~1 MB per SLC vs 4-8 GB for the full ZIP).
- `parse_relative_orbit_numbers(manifest_bytes)` — extracts `(start_track, end_track)` from `manifest.safe`.
- `parse_ascending_node_time(annotations)` — extracts `ascendingNodeTime` from the annotation XML.
- `parse_burst_sensing_times(annotations)` — extracts per-burst `sensingTime` per subswath.
- `compute_esa_burst_id(...)` — pure-Python port of `s1reader.s1_burst_id.S1BurstId.from_burst_params()` (handles ANX-crossing scenes via the published ESA formula and constants from Table 9-7).
- `derive_burst_ids_from_metadata(annotations, manifest_bytes)` — wraps the above and returns ASF-format burst IDs (e.g. `"130_279229_IW1"`).
- Older `derive_burst_ids(...)` (the buggy ASF-anchored one) is left in place but no longer called from the audit; kept for potential debugging.

### `tools/ops/cmr_audit/cmr_audit_burst_coverage.py`
- `fetch_bursts_for_slc()` rewritten to use the new annotation-only derivation. Drops the ASF SLC-BURST API call entirely.
- `_parse_asf_burst_response()` and `ASF_BURST_API_URL` removed (dead).
- Cache key bumped (`_v: 6`, plus `derivation: annotation_v1`) so pre-fix `asf_bursts` entries auto-invalidate.
- (Earlier commit on this branch) `fetch_opera_products()` now retries 429/5xx for 3 attempts and refuses to cache the result if all retries fail, so transient CMR errors no longer pollute `cmr_opera` with false-empty entries.

## Validation

- New derivation tested against `S1A_..._063027_07E8C5_7795-SLC` (the SLC where the bug was first observed). Generates 27 burst IDs that match exactly the 27 OPERA RTC products in CMR (mix of T130 + T131, due to genuine ANX crossing mid-acquisition).
- Pure-Python formula matches `s1reader.S1BurstId.from_burst_params()` constants (T_beam=2.758273, T_pre=2.299849, T_orb=12·86400/175, subswath offsets 0.832 / 1.078 / 0.848).
- No new external dependency added.

## Test plan

- [ ] Re-run Jan-Jun 2026 RTC + CSLC audits with this branch (cache invalidation will force fresh derivations).
- [ ] Diff resulting "missing SLC" batches against batches already sent (1-10) to identify previously-listed false positives we can tell OPS to skip.
- [ ] Verify `NOT caching` warnings stay correlated with real CMR degradations (not silent loss).

### DIFF
--- a/tools/ops/cmr_audit/cmr_audit_burst_coverage.py
+++ b/tools/ops/cmr_audit/cmr_audit_burst_coverage.py
@@ -758,9 +758,18 @@ async def fetch_opera_products(
                     headers={"Content-Type": "application/x-www-form-urlencoded"},
                     timeout=aiohttp.ClientTimeout(total=30),
                 ) as resp:
-                    if resp.status == 429:
-                        await asyncio.sleep(0.5 * (2 ** attempt))
-                        continue
+                    if resp.status == 429 or resp.status >= 500:
+                        if attempt < 2:
+                            await asyncio.sleep(0.5 * (2 ** attempt))
+                            continue
+                        # All retries exhausted on 429/5xx - return WITHOUT caching
+                        # so we don't poison the cache with false-empty results
+                        logger = logging.getLogger(__name__)
+                        logger.warning(
+                            f"CMR returned status {resp.status} for burst "
+                            f"{burst.filename_pattern} after 3 attempts - NOT caching"
+                        )
+                        return set()
                     if resp.status != 200:
                         return set()
 
@@ -769,10 +778,16 @@ async def fetch_opera_products(
                     cache.set("cmr_opera", cache_params, product_ids)
                     return set(product_ids)
 
-            except (asyncio.TimeoutError, aiohttp.ClientError):
+            except (asyncio.TimeoutError, aiohttp.ClientError) as exc:
                 if attempt < 2:
                     await asyncio.sleep(0.5 * (2 ** attempt))
                     continue
+                # All retries exhausted - return WITHOUT caching
+                logger = logging.getLogger(__name__)
+                logger.warning(
+                    f"CMR network error for burst {burst.filename_pattern} "
+                    f"after 3 attempts: {exc} - NOT caching"
+                )
                 return set()
 
     return set()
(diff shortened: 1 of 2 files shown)


stirlingalgermissen (approved):
Looks great 👏👏👏

pymonger:
@hhlee445 @stirlingalgermissen: i'm rerunning the RTC and CSLC audit using this fix for 2026 January to June. It will take approximately 10 days to rebuild the CMR burst cache based on just the annotation XML's only (no longer using ASF burst API).

hhlee445 (approved):
LGTM

PR #1464, opened 2026-07-22
Create SLC S1D query timer and associated jobspecs

hhlee445:
Add SLC S1D timer to trigger RTC-S1/CSLC-S1 with S1D data


RKuttruff (commented):
I think we can also remove the S1A timer since it's no longer needed

hhlee445:
We will keep the S1A timer for a little longer and remove it in the next release.

PR #1466, opened 2026-07-23
Add analysis-mirror incremental refresh tooling

pymonger:
## Summary

Adds tooling to keep the OPERA analysis instance current. The analysis instance is a single-node,
read-only OpenSearch mirror that serves copies of the OPS-FWD and OPS-POP1 clusters so analysts and
`opera-monitoring` can query operational data without touching production. It was provisioned and
seeded per the DSIO ticket (DSIOS-10159) requesting an OPERA OpenSearch analysis instance, but nothing kept it
up to date after the initial seed. This fills that gap by refreshing the mirror from the hourly S3
snapshots the OPS clusters already write.

Follow-up efficiency/availability work is tracked in
[OPERA-2589](https://hysds-core.atlassian.net/browse/OPERA-2589).

## Changes

New files under `tools/analysis/`:

- **`mirror_refresh.py`** (new) — incremental refresh driver. Standard library only, Python 3.6
  compatible, since the mirror host has no mozart/conda environment for analyst accounts.
  - **Venue prefixing.** Both venues live in one cluster and 162 index names collide between them
    (most operational time-series plus several `grq_` product indices that hold genuinely different
    documents per venue). Every index is restored under a `fwd_*` / `pop1_*` prefix, so a query that
    omits the venue matches nothing rather than silently returning half the picture.
  - **Change detection.** Each index is fingerprinted as `file_count:size_in_bytes` from the snapshot
    `_status` API and compared against the fingerprint recorded on its last successful restore
    (not against the previous snapshot's incremental stats). Comparing to our own last-restored
    state makes skipped refresh cycles self-healing, and a fingerprint is recorded only on success,
    so a failed index is retried next run rather than silently adopted as current.
  - **Restore.** Changed indices are deleted locally and restored directly from the snapshot, which
    preserves the snapshot's own mappings exactly — no reindex step, no dependency on registered
    index templates.
  - **Guards.** `--init-state` / `--init-snapshot` adopt state after a bulk seed without a full
    re-restore; `--max-changed` aborts a runaway re-seed from a lost state file.
- **`run_mirror_refresh.sh`** (new) — `flock`-guarded cron wrapper so a slow run cannot overlap the
  next tick.
- **`README.md`** (new) — operational doc: venue-prefix scheme, access/credentials, incremental
  refresh, full re-seed recipe, and the version-compatibility constraint.

## Documentation

- `tools/analysis/README.md` — setup, operation, and re-seed procedure for the mirror.

## Test plan

- [x] `mirror_refresh.py` compiles clean under Python 3.6 on the mirror host (stdlib only).
- [x] `--init-state` pinned to the seed snapshots adopts fingerprints without restoring
      (both venues), correctly bypassing the `--max-changed` guard.
- [x] Incremental run detects only changed indices via fingerprint delta (FWD tens/cycle, POP1 a
      couple/cycle), leaving unchanged indices untouched; cluster stays green.
- [x] A full cron-path run reports `restored=N failed=0`, including re-restore of a 100+ GB
      current-month product index.
- [x] `--max-changed` guard aborts a stateless run (FWD 306 changed exceeds the default limit)
      while a smaller venue proceeds.
- [x] `flock` wrapper skips a tick when a prior run is still in progress.

### DIFF
--- a/tools/analysis/README.md
+++ b/tools/analysis/README.md
@@ -1,0 +1,145 @@
+# OPERA analysis mirror
+
+A single-node OpenSearch instance holding read-only copies of the OPS-FWD and
+OPS-POP1 clusters, so analysts and `opera-monitoring` can query operational data
+without touching production.
+
+Data arrives via the hourly snapshots the OPS clusters already write to S3. The
+mirror registers those repositories read-only and restores from them; it never
+writes to the buckets and never connects to the OPS clusters.
+
+## Venue prefixes
+
+**Every index is prefixed with its venue — `fwd_*` or `pop1_*`. No index keeps
+its canonical HySDS name.**
+
+Both venues run a cluster named `common_cluster` and **162 index names collide**:
+most of the operational time-series (`worker_status`, `logstash`, `task_status`,
+`job_status`, `event_status`) plus several `grq_` product indices such as
+`grq_v1.0_l2_rtc_s1`, `grq_1_l1_s1_slc`, and `grq_v1.1_l2_cslc_s1`. The colliding
+`grq_` indices hold genuinely different documents per venue — forward products
+versus reprocessed ones — so merging them would be wrong, not merely untidy.
+
+Prefixing makes a venue-less query match nothing instead of silently returning
+one venue's data and looking complete:
+
+```
+one venue    fwd_job_status-*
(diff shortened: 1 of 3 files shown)


hhlee445 (approved):
LGTM

PR #1468, opened 2026-07-28
OPERA-2561: Enhance historical processing tool to support processing of datasets with large gaps

pymonger:
## Summary

Adds phased (mode-switching) historical processing for DISP-S1 frames whose acquisitions have
multi-year gaps. The consistent burst database is published in an annotated variant whose
`sensing_time_list` maps each sensing time to a processing-mode label, and the historical
processing tool drives each frame through those phases: historical blocks as k-set batch jobs,
forward blocks advanced one cursor position at a time, unprocessable blocks skipped, and a
**fresh compressed CSLC lineage at every new historical phase** — the reset the ticket calls for, achieved by bounding the
lineage computations rather than by deleting products.

The whole capability is gated on a settings master switch (`DISP_S1_PROCESSING_MODE_ENABLED`,
default `false`) plus a per-batch-proc `"phased": true` opt-in. With the switch off the labels are
dropped at load: an annotated database parses exactly like an un-annotated one, every frame comes
back without phases, and all phase-aware behavior falls through to what it does today.

Resolves [OPERA-2561](https://hysds-core.atlassian.net/browse/OPERA-2561)

## Burst database format

Both variants are published for each vintage. They differ in one structural respect: in the
annotated variant a frame's `sensing_time_list` is a mapping of sensing time to processing-mode
label rather than a list of sensing times. Frame 24726, the ticket's example frame (112 sensing
times over 27 bursts, 3.5-year gap between `2021-12-16` and `2025-05-29`):

**Un-annotated variant**

```json
{
  "metadata": {
    "generation_time": "2026-08-10 14:44:28.957509",
    "blackout_file": "opera-disp-s1-blackout-dates-2026-08-10.json",
    "input_cmr_csv": "cmr_survey_2016-07-01_to_2026-04-30.csv",
    "opera_db_file": "opera-s1-disp-0.18.0.gpkg",
    "generated_burst_db": "cslc-burst-database-2026-08-10.duckdb"
  },
  "data": {
    "24726": {
      "burst_id_list": ["t093_197801_iw1", "t093_197801_iw2", "..."],
      "sensing_time_list": [
        "2016-07-09T01:33:16",
        "2016-08-02T01:33:17",
        "...",
        "2021-08-06T01:33:06",
        "2021-08-18T01:33:07",
        "...",
        "2021-12-16T01:33:07",
        "2025-05-29T01:32:39",
        "...",
        "2026-04-30T01:32:43"
      ]
    }
  }
}
```

**Annotated variant**

```json
{
  "metadata": {
    "generation_time": "2026-08-10 14:44:28.957509",
    "blackout_file": "opera-disp-s1-blackout-dates-2026-08-10.json",
    "input_cmr_csv": "cmr_survey_2016-07-01_to_2026-04-30.csv",
    "opera_db_file": "opera-s1-disp-0.18.0.gpkg",
    "generated_burst_db": "cslc-burst-database-2026-08-10.duckdb",
    "reconciled_from": {
      "old_db": "provided",
      "new_db": "provided",
      "reconciliation_time": "2026-08-10T14:45:58.890533"
    },
    "processing_mode_params": {
      "batch_size": 15,
      "gap_threshold_years": 2.0,
      "labeling_time": "2026-08-10T14:45:59.229423"
    }
  },
  "data": {
    "24726": {
      "burst_id_list": ["t093_197801_iw1", "t093_197801_iw2", "..."],
      "sensing_time_list": {
        "2016-07-09T01:33:16": "historical_01",
        "2016-08-02T01:33:17": "historical_01",
        "...": "...",
        "2021-08-06T01:33:06": "historical_01",
        "2021-08-18T01:33:07": "forward_01",
        "...": "...",
        "2021-12-16T01:33:07": "forward_01",
        "2025-05-29T01:32:39": "historical_02",
        "...": "...",
        "2025-11-25T01:32:44": "historical_02",
        "2025-12-19T01:32:43": "forward_02",
        "...": "...",
        "2026-04-30T01:32:43": "forward_02"
      }
    }
  }
}
```

`historical_NN` blocks are whole k-set ministacks (`batch_size` above is the k they were generated
for), `forward_NN` is the sub-k remainder of the same chunk, and a third label `no_run` marks a
gap-separated chunk with too few full-coverage acquisitions to bootstrap a lineage at all. `NN`
numbers the chunk, not the kind, so a frame's first processed phase can be `historical_02`.

Two properties of the annotated variant this design rests on:

- **Consumers see the same sensing times either way.** `parse_sensing_time_list` accepts both
  shapes and returns the same ordered sensing times from each, so a frame loaded from the annotated
  file carries exactly the sensing times it carries from the un-annotated one. Only mode-aware code
  looks at the labels.
- **It is otherwise identical to the un-annotated variant of the same vintage.** A frame-by-frame
  diff of the two `2026-08-10` files: the same 1421 frames, zero frames with a differing
  `burst_id_list`, and zero frames with a differing sensing-time sequence. Together with the master
  switch defaulting to false, that makes `DISP_S1_BURST_DB_S3PATH` pointing at the annotated file a
  behavior-neutral deployment, confirmed on a cluster under Test plan below.

`DISP_S1_BURST_DB_S3PATH` therefore defaults to the annotated file of the current `2026-08-10`
vintage, whose sensing times were checked against the deployed blackout windows: 104,967 dates
across the 902 frames the blackout file and the database share, zero collisions.

### New files

- **`data_subscriber/cslc/disp_s1_phases.py`** — pure phase model: parsing of the annotated
  `sensing_time_list` (timestamp → label mapping, paired by timestamp so labels survive sorting),
  validated segmentation into contiguous `historical_NN` / `forward_NN` / `no_run` phases, and the
  lineage-boundary helpers. No settings, no I/O, no ES.

### Enhancements

- **`conf/settings.yaml`** — `DISP_S1_PROCESSING_MODE_ENABLED` master switch, default `false`.
- **`data_subscriber/cslc_utils.py`** — `process_disp_frame_burst_hist` reads the annotated format
  and hangs the derived phases off each frame, gated on the switch. A frame whose labels violate
  the labeler's contract keeps `phases` unset and records why in `phase_error`, so one bad frame is
  quarantined instead of failing a load every DISP-S1 component depends on. Phase-aware
  `calculate_historical_progress` (historical phases in whole k-sets, forward per date, `no_run`
  out of scope). Shared batch-proc frame expansion and batch-proc validation helpers.
- **`data_subscriber/cslc/cslc_dependency.py`** — every dependency computation counts from the
  start of the lineage containing the date being processed rather than from the beginning of the
  series. Without annotations that start is position 0, which is the existing behavior exactly.
  Lineage-relative positions are what make a post-gap block work at all: such a block begins at a
  position that is generally not a multiple of k, so on an absolute grid `determine_k_cycle` cannot
  return 0 for its boundary date — no compressed CSLC is ever published and every later ministack
  stalls — and `get_dependent_ccslc_index` resolves to a boundary that will never exist.
- **`data_subscriber/cslc/disp_s1_k_cycle_evaluator.py`** — a KSC whose sensing date falls in a
  historical (or `no_run`) phase is written for bookkeeping and marked
  `superseded_by=historical_processing`, which the existing SCIFLO trigger rule already skips. That
  closes the duplicate-product window in which a multi-ministack historical block's mid-stack dates
  would fire forward SCIFLOs racing the batch job. Compressed CSLC selection, the k-boundary anchor
  and its position fallback, the k-window date universe, the partial-lineage scan and the
  pending-boundary projection are all bounded by the historical phase that starts the containing
  chunk's lineage — for a date in a `forward_NN` block, that is the `historical_NN` block ahead of
  it in the same chunk.
- **`data_subscriber/cslc/disp_s1_constants.py`** — `SUPERSEDED_BY_HISTORICAL_PROCESSING`.
- **`tools/run_disp_s1_historical_processing.py`** — the phase walk: per-frame planning of one
  action per poll (submit a k-set query, submit a forward date, record a terminal disposition, skip
  an unprocessable block, quarantine), forward dates driven one at a time through a narrow-window
  `cslc_catalog_ingest` and advanced only once the date's k-cycle state config reaches a terminal
  disposition (fired, fired at a boundary, superseded, or no-fire on an unresolved gap), and all
  per-frame
  state (`frame_phases`, `forward_inflight`, `lineage_transitions`, `quarantined_frames`,
  `stalled_frames`) persisted on the batch proc document so a killed daemon resumes without
  resubmitting or skipping work.
- **`tools/pcm_batch.py`** — batch proc validation at creation, in both directions:
  - a **phased** batch proc is rejected when its frames are not annotated, when their annotations
    would be quarantined, or when its `k` differs from the batch size the labels were generated for;
  - an **un-phased** batch proc is rejected when a k-set it would actually submit straddles a
    gap-separated chunk boundary, naming the k-set and the two phases it spans. Stepping k dates at
    a time from the start of the series ignores the labels, so such a stack puts a multi-year gap
    inside one ministack, which the SAS rejects outright — the whole k-set is wasted compute and
    the failure would otherwise surface only after a query, a download and a SCIFLO have run.

    The test is the chunk ordinal, not the label: a `historical_NN` → `forward_NN` transition is a
    label change within one chunk, with no gap between the dates either side of it, and the SAS
    processes such a stack happily. Only the k-sets the batch proc would submit are checked, so an
    un-phased run whose date range stops short of a chunk boundary passes. Because the check reads
    the labels it is inert until the master switch is on, so what it guards in practice is the
    mistake of omitting `"phased": true` during a campaign that has the feature enabled.

    One case it cannot catch: consecutive unusable chunks are merged into a single `no_run` run,
    so a k-set falling entirely inside one may still straddle a gap.

  Validation expands a frame range inclusively of its end frame, matching what the walk does at
  runtime.

## Operator tooling and accountability

The annotated burst database is the **source of truth for what a campaign owes**: its labels fix
which jobs will be submitted and which products must exist when the campaign is finished, and that
expectation is fixed before a single job runs. Accountability is reconciliation of that expectation
against two live sources — **product existence** and **job status**. Nothing is inferred from the
batch proc cursor, which advances on *submission* and so credits work that has not finished.

### New

- **`tools/purge_superseded_cslc_granules.py`** — removes catalog documents superseded by an ASF
  reprocessing. Processing no longer depends on this being run (see *Reprocessed CSLC granules*
  below), so it is hygiene rather than a fix: the superseded documents still inflate every count
  taken off the catalog — granule audits, coverage reconciliation against CMR, anything comparing
  catalog totals to DAAC totals. Only the GRQ document is removed; the granule lives in the ASF
  DAAC bucket and is never touched.

- **`tools/disp_s1_campaign_status.py`** — the accountability view. Enumerates every unit the burst
  database owes (one per k-set in a `historical_NN` phase, one per date in a `forward_NN` phase,
  none for `no_run`), then settles each against products and jobs.

  Job attribution is exact rather than by frame: every DISP-S1 job name carries the frame, and all
  but one carry enough to pin the job to a single unit — `cslc_query_hist` carries the k-set's date
  span, `cslc_catalog_ingest` the forward date, and the SCIFLOs and `cslc_download` carry
  `latest_acq_index`, which maps back to a sensing position through
  `determine_acquisition_cycle_cslc`. Each unit takes the worst state among its jobs, with product
  existence overriding — so a retry that publishes clears the failure with no operator bookkeeping.

  **It keeps reporting after the campaign finishes.** A campaign disables its batch proc the
  moment it completes, so selecting campaigns on `enabled` reported nothing from the point the work
  was done — precisely when an operator is checking what was produced and whether anything was left
  owed. Campaigns are selected on whether they were ever given frames, most recently active first.
  A campaign that produced everything it owed with nothing failed or missing collapses to one line
  so several finished campaigns cannot bury the one still needing attention; `--all` forces detail
  and `--id` always wins.

  Three things this makes visible that counts alone cannot:
  - a unit whose job **failed** is reported as failed rather than sitting at `pending` forever,
    with the failing job id printed for a retry;
  - a job that reads `job-completed` but produced no products is a failure, not progress;
  - a frame whose **earliest outstanding unit** failed is flagged `*** STUCK ***`, because
    everything behind it gates on a compressed CSLC that will never publish and retrying a later
    k-set cannot help.

  `--failures` drops every healthy frame; `--json` feeds the plotter; the exit code is 2 when a
  frame is stuck, 1 when there are non-blocking failures and 0 when clean, so it can run as a
  scheduled campaign health check.

  **STUCK is claimed only for a historical k-set.** Its compressed CSLC gates every later k-set, so
  a failure there stops the frame and retrying a later k-set cannot help. Forward dates do not gate
  each other — the walk advances when a date's k-cycle state config reaches a terminal disposition —
  so a failed forward date leaves a hole in the products and the frame carries on. Still reported
  as FAILED, still needs an operator, but not STUCK.

  Per-date state additionally distinguishes a lineage **start** from a lineage **reset**. A
  compressed CSLC lineage begins at the first date of *every* historical phase — `lineage_start_pos()`
  returns `phase.start_pos` unconditionally, `cslc_dependency` slices and re-zeros the k-cycle
  counter from it, and `expected_units()` already withholds a product from that date. Only the
  post-gap ones also *reset* one, which is what `is_new_lineage` means and what
  `lineage_transitions` records. Keying the marker off `is_new_lineage` drew nothing at all on a
  frame whose timeline opens with `no_run`, where the first historical phase has a real lineage
  bound the runtime honours while the flag is `False`.

  A k-set is counted done only once its compressed CSLC boundary has published, not when its
  products appear — a k-set with all its products and no boundary is exactly the state that stalls
  the frame behind it.

- **`tools/plot_disp_s1_campaign_status.py`** — renders that JSON as the same picture the burst
  database gives, with accountability painted on top. Four lanes: sensing ticks coloured by state;
  a **job row** carrying one mark per job — a bracket over a k-set's 15 dates, since that is a
  single `SCIFLO_L3_DISP_S1_hist`, plus a constant-size chip at every job's end; the compressed
  CSLC boundary each k-set publishes, filled once it exists; and a **blackout rail** showing the
  seasons the labeler excludes. Multi-year acquisition gaps are hatched, so the two reasons a
  stretch of axis is empty are never confused. matplotlib only, so it runs off-cluster.

  Deliberately *not* a filled time-extent bar for a k-set: frame 33065's second k-set spans three
  years of wall clock, because snow blackout suppresses acquisition for most of each year, while
  the job runs 7-9.5 hours. Bracket ends sit on the k-set's true first and last sensing date
  so they line up with the ticks above; the chip terminates the right end, because a cap *and* a
  chip both fell inside the ~7 px that separates consecutive k-sets and merged into one mark.

  Two colour defects were fixed here that predate the job row. `done` and `failed` were an
  equiluminant green/red pair at 1.06:1, so the one distinction the figure exists to make did not
  survive deuteranopia (~8% of men), greyscale print or a washed-out projector; they are now
  separated by luminance as well as hue, with failure the darkest and thickest mark plus an ✕ and
  a chip. The lineage and boundary markers were a warm orange ~3 OKLab units from the failure
  colour; they are structure, not status, and are drawn in ink.

  Blackout windows come from `localize_disp_blackout_dates()`, clipped per frame, and are a bottom
  rail rather than full-height bands — on a snow frame the windows cover ~65% of the axis, and
  full-height bands invert figure and ground until the white gaps read as the bands. The status
  JSON carries the burst database and blackout filenames, and says so explicitly when the blackout
  file could not be read, since otherwise "no windows" and "no file" render identically and some
  frames legitimately have neither.

- **`tools/derive_phased_frame_states.py`** — emits a `frame_states` map for a new batch proc from
  the `batch_proc` index, which is the authoritative record of what previous campaigns submitted.
  Takes the furthest cursor any previous proc reached per frame, names which proc that was so a
  stale cursor from an abandoned campaign cannot silently win, and **rewinds a legacy cursor that
  is not phase-aligned**.

- **`tools/reconstruct_phased_frame_states.py`** — rebuilds `frame_states` from the compressed CSLC
  catalog for the case where `batch_proc` was not snapshotted or not restored onto a new cluster. A
  boundary is published at the last date of a completed k-set, so the cursor is the position just
  past a frame's newest boundary. Conservative by construction: it reports what *completed* rather
  than what was *submitted*, so a frame that had moved into a forward block is rewound and those
  dates are re-submitted. It never skips work.

- **`tools/validate_phased_frame_states.py`** — validates a batch proc file's cursors against the
  deployed burst database before creation. A cursor inside a `historical_NN` phase must satisfy
  `(cursor - phase.start_pos) % k == 0`; a misaligned one is not a clean failure at runtime — the
  walk submits an offset k-set, runs past the end of the phase, and quarantines the frame with a
  message that blames the phase rather than the cursor.

### Made phase-aware

Seven pre-existing DISP-S1 operator tools computed against the absolute k-set grid from position 0.
Two were silently destructive, two reported wrong numbers, and one submitted work the pipeline
discards.

- **`tools/ops/disp_s1_status/disp_s1_hist_status.py`** — its denominator was
  `len(sensing_dates) // k * k`, which assumes every date is triggerable in whole k-sets. The
  cursor also advances past `forward` and `no_run` dates, so a *completed* frame 17235 run reported
  **140%**. Now derives the denominator from the phases through the same helper the daemon uses:
  17235 140% → 100%, 33065 117% → 100%, 24726 109% → 100%, with progress monotonic and capped.
- **`tools/disp_s1_k_cycle_date_analyzer.py`** — computes the `frame_states` cursor for an end
  date, and walked k-sets on the absolute grid. For end date 2026-01-15 it emitted cursor 90 for
  frame 24726 and 45 for 33065; both land inside a `historical_02` misaligned, which is precisely
  the failure the validator above exists to catch. Gated behind `--use-processing-modes`.
- **`tools/disp_s1_burst_db_tool.py`** — only the `frame` subcommand's k-cycle grouping was wrong;
  it folded `forward` and `no_run` dates into k-sets that will never be submitted together.
- **`tools/analyze_disp_s1_forward_processing_timeline.py`** — a lineage reset read as a
  discontinuity in its timeline.
- **`tools/disp_s1_forward/submit_forward_for_frames.sh`** — submitted forward work for dates a
  historical phase owns, which the k-cycle evaluator then supersedes: wasted submissions, no
  products.
- **`tools/update_disp_s1_burst_db.py`**, **`tools/truncate_disp_s1_burst_db.py`** — both wrote
  `sensing_time_list` back as a plain *list*, silently turning an annotated database into an
  un-annotated one. They now preserve labels, validate the resulting phase layout before writing,
  and refuse rather than corrupt; `--drop-processing-modes` makes the un-annotated write explicit
  and strips `metadata.processing_mode_params` so the file cannot be mistaken for annotated.

  `update_disp_s1_burst_db.py` additionally matches sensing times **by date, not by timestamp**.
  CMR and the labeler pick different representative datetimes for the same acquisition cycle —
  differing by seconds — so timestamp comparison made every date look new and the tool refused on
  frames it could have updated safely.

## Documentation

- **`tools/README.md`** — operator section for phased historical processing: the annotated database
  format, both switches and the venue-wide scope of the master switch, a batch proc template, what
  the daemon does per phase, the state fields to read when a frame is quarantined or stalled, and a
  table of the setup and accountability tools above.
- **[How-To DISP-S1 Phased Historical Processing (Large-Gap Frames)](https://wiki.jpl.nasa.gov/spaces/operasds/pages/1310031172/)**
  — the OPS procedure, under Procedures. Prerequisites, standing a campaign up on a newly deployed
  cluster, the regional batch proc, campaign status and accountability, migrating a running
  campaign to phased, and troubleshooting. Every command on it was run against a live cluster
  before it was published.

## Test plan

- [x] **207 new unit tests, all under `tests/unit/` so CI runs them** — phase model, parser and
      feature gating, phase-aware dependency math, evaluator guards, burst-database exclusions,
      CSLC de-duplication, and the historical processing tool's state machine. 56 of that total are
      the operator tooling tests broken out below.
- [x] Golden masters: with the master switch off, the annotated database parses and drives job
      submission identically to the un-annotated one
- [x] Full simulated lifecycle through `proc_once` — four-phase frame and the real annotations of
      frame 24726 — asserting the exact job sequence, `--m=1` on the post-gap k-set, a single
      lineage transition, and no duplicate submissions
- [x] Crash/restart and lagging-cascade simulations: no resubmission, no skipped date
- [x] Smoke-test coverage for the phase walk. **Runtime: `dev-e2e-pge-DISP_S1` takes roughly
      7 hours end to end** — measured on the pyoon-0044 run, terraform apply and PCM/PGE install
      complete at 23:28Z, the un-phased stage's datasets verified at 01:43Z (~2h15m), and the added
      phased stage verified at 05:08Z (~3h25m more). The phased stage runs a single k-set on a
      1-burst frame, so it is the cheapest possible exercise of the walk; budget it as **+3.5 h on
      top of the existing test**, and note the deploy itself is a further ~1.5 h before the first
      stage starts. `dev-e2e-pge-DISP_S1` gains a phased stage after its
      un-phased one (that ordering is load-bearing — the un-phased stage is the behavior-neutrality
      check and must observe the switch off), walking single-burst frame 16936 over
      `historical_01`'s first k-set. The master switch is venue-wide and is restored before the
      forward stage, whose window on frame 31241 lies inside that frame's `historical_01` block and
      would otherwise be superseded wholesale. Stage expectations are scoped by
      `metadata.frame_id`, since `check_datasets_file.py` treats a count as exact and gives up as
      soon as it is exceeded.
- [x] `dev-e2e-pge-DISP_S1-all_phases`, a new periodic cluster for DISP-S1 pipeline changes: frame
      17235, the cheapest shipping frame carrying every phase kind — `historical_01[15]`,
      `forward_01[1]`, a 2.8 year gap, `historical_02[15]`, `forward_02[2]`, a 3.8 year gap,
      `no_run[9]`. Its `data_end_date` runs past the trailing `no_run` so the skip path executes
      rather than the walk ending early. `check_disp_s1_phases.py` additionally asserts the compressed CSLC
      boundaries fall at phase-relative positions, that `no_run` dates produced nothing, and that
      every historical-phase KSC is superseded — counts alone cannot distinguish a correct phased
      walk from a regression that reverted to the absolute grid.

      **Runtime: budget ~24 hours end to end, and set `--max_time` accordingly (the cluster ships
      with 108000 s / 30 h).** Frame 17235 is 27 bursts, and 27-burst historical k-sets were
      measured at **8.0 h and 8.5 h** on this very frame. The critical path is
      `historical_01`'s k-set (~8.5 h) → `forward_01` → `historical_02`'s k-set (~8.5 h, a fresh
      dependency-free lineage) → `forward_02` → the trailing `no_run` skip, plus ~1.5 h of
      terraform apply and PCM/PGE install before any of it starts. That is why this is a periodic
      cluster and not in the deploy path: it cannot fit in a PR gate.
- [x] **Those 56 operator-tooling tests, broken out.** They sit under `tests/unit/tools/`.
      `pytest.ini` sets `testpaths` to `tests/unit` and `docker/run_tests.sh` runs a bare `pytest`,
      so nothing under `tests/tools/` has ever executed in CI; both modules were moved there rather
      than left silently unrun. **41** drive the accountability reconciliation directly
      (`test_disp_s1_campaign_status.py`) so the failure paths a healthy cluster never exercises
      are covered — a failed job makes a unit failed rather than pending; a `job-completed` with no
      products is a failure; products win over a stale failure so a successful retry self-clears;
      a failure on the earliest outstanding unit is stuck, one behind a running unit is not yet;
      `no_run` is never failed or pending; partial output inside a failed k-set still shows the
      dates that published; and a campaign that has completed and disabled itself still reports.
      **15** pin the phased k-cycle analyzer to phase-relative k-sets
      (`range(phase.start_pos, phase.end_pos, k)`), including the fallback for an un-annotated
      frame inside an annotated database.
- [x] **Mid-campaign cutover to a regional batch proc.** Three per-frame batch procs were
      replaced by a single phased regional proc on a dev cluster while all three frames were part
      way through, each in a different phase kind at the moment of cutover — 25278 and 24726
      mid-`historical_01`, 33065 in `forward_01`. The walk resumed at the seeded cursors and
      correctly waited on the k-sets that were in flight rather than resubmitting them. Expected
      counts derived from the burst database reconciled exactly against the phase structure:
      24726 113 products / 31 jobs, 25278 58 / 4, 33065 51 / 11 — 222 products and 263 compressed
      CSLCs for the region.
- [x] **Frame 25278 (Okmok) — 58/58 products.**
      Counts match what the burst database owes exactly: `historical_01[45]` → 44 products +
      `historical_02[15]` → 14 = 58, and 4 k-sets × 14 bursts = 56 compressed CSLCs, both observed.
      58 products over 58 distinct secondary dates, no duplicates. `check_disp_s1_phases.py` passes
      every assertion on it — the four compressed CSLC boundaries fall at phase-relative positions
      (`20180611`, `20191004`, `20211017` in `historical_01`; `20260730` in `historical_02`, which
      starts at position 45 and so could not land on the absolute k-grid), and every
      historical-phase KSC carries `superseded_by=historical_processing`. That is the check counts
      alone cannot make: it distinguishes a correct phased walk from a regression to the absolute
      grid.
- [x] **Frame 33065 (Unimak) — 51/51 products**, and it is the trap case:
      `historical_01[30]`, `forward_01[8]`, `historical_02[15]` starting at position **38**, which
      is not a multiple of k. Its post-gap boundary lands at position 52; a walk on the absolute
      k-grid would have put boundaries at 44 and 59 and never at 52, so the observed boundary set
      `20171004`, `20201012`, `20260708` could only come from a phase-relative walk. Accounting is
      exact: 29 + 8 + 14 = 51 products, 3 k-sets × 15 bursts = 45 compressed CSLCs, 51 distinct
      secondary dates, no duplicates.

      **The lineage reset is visible in the products themselves.** Grouping this frame's products
      by their reference date gives `20160723` → 14 and `20171004` → 15 (the two `historical_01`
      k-sets), then eight rolling single-product references through `forward_01`, then
      `20250531` → 14 for `historical_02` — which references *its own first date*, not anything on
      the far side of the 3.6-year gap. That is the behaviour the ticket asks for, shown in the
      output rather than asserted from the code.

- [x] **Frame 24726 (Arizona) — 113/113 products, and the region complete at 222/222 with the
      proc self-disabled.** 24726 is the hardest of the three: `historical_01[75]`,
      `forward_01[11]`, `historical_02[15]`, `forward_02[14]`, six compressed CSLC boundaries
      across two historical phases. `check_disp_s1_phases.py` passes every structural assertion on
      it — the boundary set `20170429`, `20200120`, `20200811`, `20210207`, `20210806`,
      `20251125`, every historical-phase KSC superseded, no `no_run` block skipped incorrectly, and
      `all 25 forward date(s) produced a product`, which is the assertion the whole forward path
      exists to satisfy. It passes on all three frames.

      The product count follows the phase structure exactly, and the arithmetic is worth stating
      because it is not `one product per sensing date`: the first k-set of a historical phase
      yields k−1 products, because the phase's own first date becomes the compressed CSLC lineage
      reference and has nothing to pair against. 24726 has 115 sensing dates and two historical
      phases, so 74 (`historical_01`) + 11 (`forward_01`) + 14 (`historical_02`, a single k-set and
      therefore a first) + 14 (`forward_02`) = 113. Same shape on the others: 25278 44 + 14 = 58,
      33065 29 + 8 + 14 = 51.

      One product in that total needed a recovery, and it was **not** a phased-processing defect.
      Frame 24726's `20260605` was blocked by an ASF reprocessing that put two CSLC granules in the
      catalog for the same burst and sensing date — the failure mode fixed by *Reprocessed CSLC
      granules on the forward path* below. It published on re-drive under the de-duplicated list,
      taking the frame to 113/113 and the region to 100%.
- [x] **The accountability tool found two defects in itself on that live campaign**, both of which
      reported healthy work as broken, and both now covered by tests: the forward SCIFLO's job name
      matched none of the attribution patterns, so a running forward date was reported FAILED; and
      "every job completed but no products" fired on a completed *upstream* job, when it should
      only fire once a job that actually emits products has completed. Fixing them is the reason
      the exit code now distinguishes stuck from non-blocking failures.
- [x] Container build: 487 passed. The 21 failures are environmental — `assertEquals` removed in
      Py3.12, a missing MGRS track-frame DB, `KeyError: 'runconfig'` in product2dataset, and a test
      that queries live CMR — and none is in a file this branch modifies or that imports a module
      it touches.
- [x] **Behavior-neutral deployment**, dev cluster with the annotated database as the shipping
      default and the master switch off: the stock DISP-S1 smoke test produced 14 `L3_DISP_S1` +
      2 `L2_CSLC_S1_COMPRESSED`, and the stock un-phased batch proc passed validation
      (`This batch_proc seems valid`) against the annotated database.
- [x] **Full phased lifecycle**, dev cluster, master switch on, real PGE execution, on a
      purpose-built database fixture: the shipping 1421-frame file with frame 31241's labels
      replaced by `no_run[29] h01[30] f01[4] h02[15] f02[4] no_run`. The substitution is
      deliberate — 31241 carries only 2 bursts, which keeps real-PGE execution affordable, while no
      2-burst frame in the shipping database happens to exercise every label kind. Of 1421 frames,
      116 have more than one historical chunk (a real multi-year gap) and 3 exercise every label
      kind; the smallest, frame 17235
      (`h01[15] f01[1]` / 2.8 yr gap / `h02[15] f02[2]` / 3.8 yr gap / `no_run[9]`), carries 27
      bursts:
  - job sequence exactly as designed — `no_run` skipped whole, two `cslc_query_hist` for
    `historical_01`, four serial ingests for `forward_01`, one `cslc_query_hist` for
    `historical_02` with `--m=1`, four serial ingests for `forward_02`, trailing `no_run` skipped,
    batch proc self-disabled at 100%
  - compressed CSLC boundaries `20200619`, `20210602` (`historical_01`, one per k-set) and
    `20220727` (`historical_02`, which starts at position 63 — not a multiple of k)
  - all 30 historical-phase KSCs carry `superseded_by=historical_processing`; exactly 8 forward
    SCIFLOs ran, one per forward date, no duplicates
  - `forward_02` KSCs reference only `20220727`, ignoring the three older boundaries in GRQ —
    `20200619` and `20210602` from this frame's `historical_01` lineage, and `20171021` from the
    smoke test that ran on the same frame beforehand — with the evaluator running m=6
  - 65 L3 products over 65 distinct dates, no duplicates: 51 from the phased run (its 53 processed
    dates less the two lineage reference dates, which yield no displacement product) and 14 left by
    the stock smoke test, which runs this same frame through 2017-10-22
  - exactly one `lineage_transitions` entry, for `historical_02`

## Reprocessed CSLC granules on the forward path

ASF republishes a burst under a new processing date when it reprocesses it. Both granules stay in
the catalog, both get staged, and the SAS refuses a stack containing two granules for the same
burst and sensing date — `disp_s1/main.py::_assert_no_duplicate_dates`. So a single reprocessing
event silently blocks every affected sensing date, and it surfaces as a crash rather than a
rejection: the guard fires correctly and then dies building its own error message, which joins
`Path` objects with a string separator (`TypeError: sequence item 0: expected str instance,
PosixPath found`). The guard is right; the message is broken. Reported to ADT separately.

Observed on frame 24726: 14 of 27 bursts on `2026-06-05` carried both a `2026-06-06` and a
`2026-07-18` granule, and that one forward SCIFLO failed while the other 24 dates ran.

`cslc_utils.latest_cslc_per_burst()` keeps one granule per burst **and sensing time** — the most
recently processed — and is applied at both points a CSLC list is built:

- `disp_s1_cycle_evaluator.py` when the cycle state config is written, which is also what the
  per-date `window_entries` lists are built from;
- `disp_s1_k_cycle_evaluator.py` when the k-cycle state config's `product_paths` are assembled,
  so a state config written by an older release still yields a clean stack.

De-duplication is per burst *and* per acquisition, never across dates, so it cannot collapse two
genuine acquisitions of the same burst. Paths that do not parse are passed through untouched
rather than dropped — the function never silently discards an input it cannot reason about.

One property worth stating because it cost a repeat failure in practice: **a state config freezes
its file list at build time.** Purging the catalog does not rewrite lists already written, so a
date re-driven after a catalog purge re-submits the frozen list and fails again — and the
superseded paths then point at documents that were just deleted. The code above prevents new
occurrences; a cluster carrying state configs built before it must have those repaired, in all
three places the list is stored (`cslc_product_paths` on the cycle state config, and both
`product_paths.L2_CSLC_S1` and `window_entries[N].cslc_product_paths` on the k-cycle state
config). The operator guide documents that repair.

## Burst-database exclusions in forward processing

An acquisition that covers only part of a frame is published to CMR anyway, and the cascade
writes a cycle state config for it. The consistent burst database leaves such a date out of
`sensing_time_list` — that omission is the database's judgement, made by the same CMR survey
that decided everything else in the file — but nothing downstream knew it. The partial state
config **occupied a k-slot** *and* **counted as an unresolved gap in the lineage**, two
independent blockers, so every later forward date of the frame stalled. Silently: a no-fire
disposition is terminal, so the walk advanced past each date and the frame reached a 100%
cursor owing every one of its forward products.

Measured on the opera-sds#158 campaign, frame 24726: **25 forward dates, 0 products, 0
failures recorded anywhere**. The blocking acquisitions were confirmed genuinely partial in
CMR — 6 of 27 bursts, an alternating track-093 pattern — so ADT was right to exclude them,
and the 25 dates it did list were certified full coverage. Both blockers had to clear: a fix
to the gap check alone would have left `cycles_complete` at 14/15 and produced nothing.

**The bound is what makes this safe.** Absence only means "deliberately excluded" inside the
range the survey assessed, read from `metadata.input_cmr_csv` and nowhere else that survives
— both databases the campaign ran on had been renamed on disk, and `generation_time` runs
weeks past the survey end. Past the assessed range the survey never looked, absence says
nothing, and the date stays ordinary forward work; without that clause the fix would silence
new acquisitions as a database ages.

Exclusion is **computed from the deployed database at read time**, never read off the
document. A newer database that lists a previously-absent date is honoured immediately, and
state configs written before this ships need no migration. `db_excluded` /
`db_excluded_reason` / `burst_db_id` are recorded for accountability only, and are set
**blackout-first**: blacked-out dates are also absent from `sensing_time_list` (71,556 dates
across 921 frames), so testing absence alone would file every snow-season acquisition as a
partial-coverage exclusion and put a false reason into the record used to reconcile with ADT.

Every failure path fails open — no metadata block, unparseable survey name, empty sensing
list, unknown frame — because the rule is never to guess a range.

`DISP_S1_BURST_DB_EXCLUSION_ENABLED` **defaults on**, including when the key is absent
entirely, so a venue running new code against an un-updated `settings.yaml` still gets the
fix. This is a correctness repair rather than an opt-in feature, and the historical path
already honours these exclusions implicitly by building its k-set grid from
`sensing_time_list` — so defaulting on is also what makes forward and historical agree on
what an acquisition *is*. The switch exists to roll back without a redeploy. Note this path
runs for **all** DISP-S1 forward processing, not only phased campaigns.

Deliberately **out of scope**, still open: `_query_cslc_catalog` certifies any date with no
state config as complete with whatever bursts happen to be present; and seven frames whose
burst lists shrank between gpkg 0.15.0 and 0.18.0 carry permanently `expected>found` state
configs on dates the database *does* list.

## Operational note: measured k-set runtime

A historical k-set is one `SCIFLO_L3_DISP_S1_hist` job that emits **all** its products at the very
end, so the product count sits flat for the whole run and then jumps by k. Measured across the 13
k-sets completed across four frames:

| Bursts in the frame | k-set runtime |
| --- | --- |
| 14 (frame 25278) | 7.1 – 7.8 h, mean 7.5 |
| 15 (frame 33065) | 7.1 – 7.7 h, mean 7.3 |
| 27 (frames 17235, 24726) | 8.0 – 9.4 h, mean 8.7 |

Burst count matters more than an early two-sample read suggested — an earlier revision of this PR
and of the wiki page said "7 to 8.5 hours, and burst count barely changes it", which understated
the 27-burst case by roughly an hour and a half. Budget **~7.5 h for a 14–15 burst frame and up to
9.5 h for a 27-burst one**.

The operational consequence is that **a flat product count is not evidence of a stall**: on this
campaign L3 sat unchanged across five consecutive two-hourly health checks while a single 9.4 h
k-set was running normally. Progress has to be judged from job status, not from product counts —
which is why `disp_s1_campaign_status.py` reconciles against both, and why the campaign health
check now records the running-job count and only flags a stall when nothing is running *and*
nothing has published.

## Operational note: what actually runs concurrently

Two concurrency rules that the batch proc fields make it easy to guess wrong, both observed on the
live campaign and now stated in `tools/README.md` and on the wiki page:

- **k-sets are sequential within a lineage, not across a gap.** Each k-set gates on the previous
  one's compressed CSLC — but the first k-set of a post-gap `historical_NN` phase starts a *fresh*
  lineage, is dependency-free, and runs alongside the tail of the phase before it. Frame 25278 ran
  the last k-set of `historical_01` and the only k-set of `historical_02` at the same time. A
  frame's long pole is its longest single historical phase, not its total k-set count.
- **`forward_inflight` is not a concurrency limit.** A forward date leaves the walk as soon as its
  k-cycle state config reaches a terminal disposition, and *firing is terminal* — the SCIFLO it
  triggered is never waited for. When the CSLCs are already in hand the dispositions resolve fast
  enough that a whole forward block is submitted in one poll cycle: frame 33065 put all eight of
  its `forward_01` dates in flight at once, with `forward_inflight` reading `null` because nothing
  was awaiting a disposition. Forward throughput is set by the SCIFLO worker pool, not by the walk.

## Operational note: pin a database vintage for the duration of a campaign

A phase boundary is a function of a date's position within its gap-separated chunk, so boundaries
are stable only for a fixed database vintage. A later CMR survey can backfill an acquisition
*earlier* in a frame's series, which shifts every subsequent date by one position and moves the
modulo-`k` boundaries with it.

This is measurable between the `2026-04-25` and `2026-08-10` vintages. No frame loses sensing
dates, but 160 frames have a shared date whose label differs. In 109 the frontier advanced — a trailing
`forward_NN` or `no_run` block reached 15 dates and became `historical_NN`. In the other 51 a label
moved the other way, `historical_NN` to `forward_NN`: frame 1086 gained a `2026-02-17` acquisition
and frame 44325 a `2026-01-05`, each landing before the date whose label changed.

The labeler is correct in both directions, but swapping vintages under a frame that is
mid-historical-run can move a k-set boundary and misalign a compressed CSLC lineage already built.
Pin the vintage for the duration of a campaign and re-plan deliberately when adopting a new one.


hhlee445:
Getting an error for frame 24718 


https://100.104.40.95/hysds_ui/figaro?resource=%22job%22&status=%22job-failed%22

pymonger:
@hhlee445 — the branch has moved a long way since your run, so the manual re-run I asked for earlier is no longer the right test. The capability is now covered end to end by a dedicated smoke test, and it is self-contained: you deploy the stack and it does the rest, so there is no batch proc to hand-build and no settings toggles to remember (the two things that made the earlier run take the legacy path).

## Running it

Deploy `cluster_provisioning/dev-e2e-pge-DISP_S1-all_phases` with `pcm_branch` set to this branch. The deploy itself runs the test — `run_smoke_test.sh` fires from the mozart provisioner, creates the batch proc, starts the historical processing daemon, waits on the expected datasets, then runs the phase-structure assertions and cleans up the daemon.

**Budget most of a day.** Frame 17235 carries 27 bursts and this runs the real PGE. The first full run took ~21.5h end to end — two k-sets plus three forward dates, each SCIFLO around 3h. The dataset wait is set to 30h so a slower PGE or a scale-up delay does not turn into a false red.

**Use your own branch for `pcm_branch` if you start while another deploy is up.** Two stacks sharing a `pcm_branch` share a container tag (`container-nasa_opera-sds-pcm:<branch>`), a Jenkins job name, and a job-type suffix, so they can overwrite each other's build mid-flight — and the failure is quiet rather than loud. Cut `OPERA-2561-<venue>` off this branch and point `pcm_branch` at that. I have one running on `collinss` against `OPERA-2561` at the moment.

## What it exercises

Frame 17235's annotations in the shipping database carry every phase kind and two multi-year gaps:

```
historical_01[15]  2016-07-07 .. 2017-09-06     one whole k-set
forward_01[1]      2017-09-18
---- 2.8 year gap ----
historical_02[15]  2020-06-22 .. 2021-08-04     one whole k-set, fresh lineage
forward_02[2]      2021-08-28 .. 2021-09-09
---- 3.8 year gap ----
no_run[9]          2025-06-20 .. 2025-09-24     skipped whole
```

That covers what the standard DISP-S1 smoke test cannot: a compressed CSLC lineage reset at a new historical phase, a boundary at a position that is not a multiple of k on the absolute grid, `--m=1` on the post-gap k-set, forward KSCs referencing only the current lineage's boundary while the pre-gap ones are still sitting in GRQ, and a `no_run` block skipped whole with the batch proc still reaching 100%.

## What to look at when it finishes

Two files on mozart, both printed at the end of the run:

- `/tmp/datasets_all_phases.txt` — the counts. 31 `L3_DISP_S1` and 54 `L2_CSLC_S1_COMPRESSED` for frame 17235.
- `/tmp/phases_all_phases.txt` — `check_disp_s1_phases.py`, which is the part worth reading closely. Counts alone do not prove the walk took the phased path; a regression to the absolute grid could still produce 31 products. These assertions read the batch proc state and the lineage directly — the phase segmentation, the compressed CSLC boundaries, that every historical-phase KSC is superseded, and that every forward date produced a product.

`check_pcm.xml` is copied back to the terraform directory for the JUnit report.

You can also run the phase assertions by hand against any frame at any point, including mid-run:

```bash
cd ~/mozart/ops/opera-pcm
python conf/sds/files/test/check_disp_s1_phases.py --frame-id 17235 --k 15
```

For reference, the same assertions pass on a three-frame regional campaign that finished today on a dev cluster: 222/222 products across frames 24726, 25278 and 33065, on PGE 3.0.11-rc.1.0 / SAS 0.5.14. That is in the PR description under Test plan, along with the rest of the evidence.

stirlingalgermissen:
@pymonger this PR summary is too long

hhlee445 (approved):
LGTM

pymonger:
## Post-merge: `db_excluded` validated on a cluster, as a controlled A/B

The description covered this feature by unit tests and by the campaign that motivated it, but that campaign was unblocked by hand-flagging 23 cycle state configs `blackout=true` — so the feature itself had never been shown doing the job on a cluster. This closes that, and it turned out to be possible without any fixture and without touching the campaign's data.

**The subject is a real omission in the shipping burst database.** Frame 9172 carries 2 bursts, `historical_01[45]` then `forward_01[7]`. CMR holds an acquisition on **2021-08-08** with **1 of the frame's 2 bursts** — a pass covering part of the frame, which the database deliberately does not list, which is not blacked out, and which falls inside the range the CMR survey assessed. That is the production scenario in miniature: same mechanism as frame 24726, at 2 bursts instead of 27.

Found by diffing the database's `sensing_time_list` against CMR per burst across the cheap frames and separating genuine omissions from blacked-out ones. Four more frames carry the same shape (20685 with 29 such dates, 32501, 22925, 13198), all with the omission inside a historical phase rather than a forward block.

**Method.** One cluster, one frame, the shipping database unmodified. Only `DISP_S1_BURST_DB_EXCLUSION_ENABLED` changed between arms; `historical_01` was walked once and both arms shared it.

### Arm A — exclusion `false`

```
CSC cslc_s1-cycle-f9172-20210808-state-config
   is_complete False   coverage 1/2   blackout False   db_excluded False

historical_01  44/44 products
forward_01      1/7  products

20210802  complete=True   cycles=15/15  gap_unresolved=False
20210814  complete=False  cycles=14/15  gap_unresolved=True
  ... 20210826 20210907 20210919 20211001 20211013 identical

reason: K-window incomplete: 14/15 CSCs complete; gap_unresolved:
        partial CSC(s) in lineage (after CCSLC last_date 20210721): 20210808 (1/2)

window_entries:  20210802 complete=True  paths=2
                 20210808 complete=False paths=1   <- occupying a k-slot
                 20210814 complete=True  paths=2
```

Both blockers fire together — the partial acquisition consumes a k-slot **and** sets `gap_unresolved`. That is why fixing only one would have produced nothing on 24726.

### Arm B — exclusion `true`

```
20210802 .. 20211013   complete=True   cycles=15/15   gap_unresolved=False
reason: ready: 15 CSLCs + 3 CCSLCs

forward_01  7/7 products      L3 total 51  (44 historical + 7 forward)
```

`20210808` is gone from `window_entries`, which is why the count moves 14/15 → 15/15.

| | exclusion off | exclusion on |
|---|---|---|
| forward products | **1 / 7** | **7 / 7** |
| `cycles_complete` | 14/15 | 15/15 |
| `gap_unresolved` | True | False |
| L3 total | 45 | 51 |

### One property worth calling out

The CSC for `20210808` was **left exactly as arm A wrote it** — still `db_excluded=False` — and arm B excluded the acquisition correctly anyway. Exclusion is computed from the deployed database at evaluation time rather than read off the cycle state config, so **state configs written before this shipped need no migration**. Clusters carrying old state configs pick up the correct behavior as soon as the code lands.

### Not covered here

- The omission is in a **forward** block. The k-slot path inside a *historical* phase is exercised by the same mechanism but was not run; frame 20685 is the candidate if that is ever wanted.
- The 23 manual `blackout=true` flags on 24726 are still in place. This result predicts they are unnecessary, but un-flipping them re-fires the cascade on a frame under review, so it is left for after sign-off.
- Worth adding frame 9172 to `dev-e2e-pge-DISP_S1-all_phases` for regression cover — it runs concurrently with the 27-burst frame that gates that stack, so it costs almost no wall-clock. Any such check must assert its own preconditions (date absent from `sensing_time_list`, not blacked out, inside the assessed range) and fail loudly if a future database vintage lists the date, or it will keep passing while testing nothing.

Both validation clusters have been destroyed; the evidence bundle (state configs, the six pre-fix KSCs, `all_phases` output) is preserved off-cluster.

PR #1473, opened 2026-08-07
Improve exception handling and resource cleanup in checksum_util and cslc_utils

ARYA-mgc:
## What
Cleaned up a few rough edges I hit while reading through `checksum_util.py` and `cslc_utils.py`.

## Changes

**checksum_util.py**
`get_file_checksum()` was a 6-branch if-elif picking md5/sha1/sha224/sha256/sha384/sha512 by string. Swapped it for `hashlib.new(algorithm)` - works for anything hashlib supports, one less place to touch when someone wants a new algo.

**cslc_utils.py**
- Two bare `except:` (lines 60, 72) were catching everything including Ctrl+C and SystemExit. Changed to `except Exception:`.
- Line 173's bare except is specifically catching a missing `"data"` key when falling back to the old JSON format, so tightened that one to `except KeyError:` instead - more honest about what it's actually handling.
- `json.load(open(file))` at lines 172, 175, 212, 262 never closes the fd. Switched to `with open(file) as f: json.load(f)`.

## Testing
- Checked `hashlib.new(name, data)` gives identical digests to the old constructors for all 6 original types.
- `get_file_checksum()` has no callers in-repo, so the RuntimeError->ValueError change on bad input is a non-issue here.
- Manually exercised both fallback paths (S3 fail -> local, missing `data` key -> old format) to confirm they still trigger correctly after narrowing the excepts.
- No unit tests exist for either function currently - worth adding separately.

No behavior change intended anywhere here, just tightening up error handling and fixing the fd leak.

### DIFF
--- a/data_subscriber/cslc_utils.py
+++ b/data_subscriber/cslc_utils.py
@@ -57,7 +57,7 @@ def localize_disp_frame_burst_hist(settings_yaml_path=None):
 
     try:
         file = localize_anc_json("DISP_S1_BURST_DB_S3PATH", settings_yaml_path)
-    except:
+    except Exception:
         logger.warning(f"Could not download DISP-S1 burst database json from settings.yaml field DISP_S1_BURST_DB_S3PATH from S3. "
                        f"Attempting to use local copy named {DEFAULT_DISP_FRAME_BURST_DB_NAME}.")
         file = DEFAULT_DISP_FRAME_BURST_DB_NAME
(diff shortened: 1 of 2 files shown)


ARYA-mgc:
Hii! @RKuttruff sorry for tagging you directly, please review this PR when you get a chance small robustness fixes, no behavior change intended

RKuttruff (approved):

ARYA-mgc:
Hi may I know why this PR was closed? I’d like to understand the reason Thank you!

PR #1474, opened 2026-08-10
Continually retry orbit file downloading for recent SLCs when none can be found

RKuttruff:
## Purpose
We've seen frequent SLC download job failures due to being unable to find any suitable orbit file for the associated SLC.

```
util.dataspace_util.NoQueryResultsException: No results returned from parsed query results
```

This occurs when the orbit file is not yet published when the SLC is being downloaded, so this PR adds a check to continually backoff the orbit file download until the SLC has reached a certain age.

## Issues
- [OPERA-2448](https://hysds-core.atlassian.net/browse/OPERA-2448)

## Testing
- [x] Ran in dev with S1C/D timers running every 10 minutes, verified orbit files retried until available.

### DIFF
--- a/data_subscriber/asf_slc_download.py
+++ b/data_subscriber/asf_slc_download.py
@@ -27,7 +27,7 @@ from util.dataspace_util import (NoQueryResultsException,
                                  DEFAULT_DATASPACE_ENDPOINT)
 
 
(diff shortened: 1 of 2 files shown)


hhlee445 (approved):
LGTM

PR #1475, opened 2026-08-12
OPERA-2557: Integrate HySDS v6.4.3, retire the HC-633 hot-patch, and move AMIs to v6.1

pymonger:
## Summary

Integrates HySDS framework **v6.4.3**, retires the interim HC-633 hot-patch this repo has been deploying via provisioning since #1447, and moves the cluster AMIs from v6.0 to v6.1.

Resolves [OPERA-2557](https://hysds-core.atlassian.net/browse/OPERA-2557). Companion: nasa/opera-sds-lambdas#65 (`hysds-commons` bump in `batch_process`).

Component deltas: hysds 3.1.1 → 3.3.2, hysds_commons 2.1.1 → **2.4.1**, lightweight-jobs 2.0.1 → **2.1.1**, mozart 3.0.0 → 3.2.0, grq2 2.3.1 → **2.4.1**, sdscli 2.0.0 → 2.1.1, osaka 1.3.1 → 1.3.2. chimera stays at v3.0.0.

hysds 3.3.2's user-rules code is the same code validated here as the v3.1.8 backport plus a per-item transient msearch retry (`es_util.py` is byte-identical), so the trigger-evaluation behaviour is already proven at scale. lightweight-jobs v2.1.1 supersedes the patched v2.0.1.1 build, carrying both the retry-query backoff and the lock-wait hardening.

### Why v6.4.3 specifically

v6.4.2 ships a defect this integration would otherwise inherit. `hysds_commons` 2.4.0's `OpenSearchUtility._pit()` calls `create_point_in_time()`/`delete_point_in_time()`, which **opensearch-py 3.0 removed** in favour of `create_pit()`/`delete_pit()`. HC-606 dropped the `opensearch-py<3.0.0` cap in v6.4.2 without touching that call site, and the v6.4.2 venv bundles ship `opensearch_py 3.2.0` — so **every OpenSearch query returning ≥ 10,000 hits raises `AttributeError`**.

`SearchUtility.query()` switches to the PIT path exactly at that threshold; below it, `from_`/`size` paging is unaffected. That is why it is invisible on a small cluster. At OPERA scale the subscriber catalogs, the audit tools and `retrieve_disp_s1_from_grq` all cross it.

Filed as [HC-645](https://hysds-core.atlassian.net/browse/HC-645), fixed in [hysds/hysds_commons#72](https://github.com/hysds/hysds_commons/pull/72), released in **v6.4.3**:

| component | v6.4.3 change |
|---|---|
| `hysds_commons` **v2.4.1** | HC-645 — `create_pit`/`delete_pit`, query args routed through `params` |
| `grq2` **v2.4.1** | HC-645 — `update_aliases` body passed by keyword (3.x made it keyword-only) |
| `lightweight-jobs` **v2.1.1** | rebuilt against the v6.4.3 base images; no source change |

## Changes

Deliberately separable commits, so they can be reviewed, shipped or reverted independently.

### 1. Integrate HySDS v6.4.3 and retire the HC-633 hot-patch

| file | change |
|---|---|
| `setup.py` | hysds_commons v2.4.0 → **v2.4.1** (the only pinned component that moves in v6.4.3); osaka v1.3.2 and hysds v3.3.2 unchanged |
| six `cluster_provisioning` defaults | `hysds_release` v6.1.2 → **v6.4.3** (`dev-e2e-pge-DISP_S1` symlinks dev's `variables.tf`, so it is covered) |
| `modules/common/mozart.tf` | remove `swap_ops_repo()` and the factotum push block from #1447. `reinstall_hysds_compat` stays — it works around pip strict-editable mode breaking `sds -d update`, unrelated to the hot-patch |
| `modules/common/mozart.tf` | import lightweight-jobs from the framework bundle again instead of the pinned v2.0.1.1 artifactory download, failing loudly if the bundle ever ships more than one candidate |
| `dev-int/override.tf` | restore `pge_snapshots_date` to `20250514-3.1.6`; #1447 bumped the suffix to `3.1.8`, conflating it with the hysds hot-patch version. It tracks the PCM branch version and the variable is inert |

### 2. Move the cluster AMIs from v6.0 to v6.1

| | before | after |
|---|---|---|
| mozart / metrics / grq / factotum | v6.0 | v6.1 |
| autoscale | v5.4.3 | unchanged |

**This is an OpenSearch major upgrade, not a freshness bump.** OPERA runs a single 3-node OpenSearch cluster (`common_cluster`) across mozart, grq and metrics — `es_cluster_mode` defaults to `true` and all three `*_ES_PVT_IP` values are the same 3-IP list — and all three take their AMI from `default_ami_versions`, so this moves the whole cluster at once:

| | OpenSearch | redis |
|---|---|---|
| v6.0 | 2.9.0 | 7.2.1 on `/tmp/redis.sock` |
| v6.1 | **3.6.0** | 8.6.2 on `/run/redis/redis.sock` |

That makes the pairing deliberate: the v6.4.x line carries HC-606, "updates to support OpenSearch 3.x" across grq2, hysds_commons, mozart and pele, completed by HC-645 in v6.4.3. v6.1 AMIs on the old v6.1.2 framework would be the unsupported combination. `*_ES_ENGINE` derives from the AMI major and is `opensearch` on both lines, so it does not change.

`default_ami_versions` is the only place this is set — `var.ami_versions` is empty and no venue overrides it — so one edit covers every venue. `autoscale` stays at v5.4.3: it names an SSM parameter rather than an AMI.

### 3. Keep the rendered celeryconfig pointed at the socket redis actually binds

hysds 3.2.1 (HC-625) changed the celeryconfig template's `REDIS_UNIX_DOMAIN_SOCKET` default to `/run/redis/redis.sock`. sdscli renders mozart's celeryconfig from the *installed hysds tree's* template — `resolve_files_dir` only prefers `~/.sds/files` when a plain `celeryconfig.py.tmpl` exists there, and this repo ships only `.asg` / `.private_verdi` variants — so the rendered path moves with the framework bump.

That matters because `scripts/process_events.py` builds its own pool from that setting and `rpush`es **every** task and worker event through it with no error handling: a mismatch crash-loops the daemon and stops `task_status` / `worker_status` indexing.

The v6.1 AMIs bind `/run/redis/redis.sock`, so on this branch's target configuration the two already agree. The provisioner therefore probes for the socket on disk, compares it against the template default, and writes a `~/.sds/files/celeryconfig.py.tmpl` override **only when they differ** — a no-op on v6.1, a correct pin on v6.0, which keeps this commit independent of the AMI commit. Probing the filesystem rather than reading `redis.conf` matters because that config is root-only on v6.1, and redis cannot be reconfigured from provisioning anyway given how narrow `hysdsops`' sudo is.

### 4. Select cluster credentials by label instead of line number

`~/.creds` is written by the AMI's cloud-init and was read positionally — rabbitmq from line 1, redis from line 2. What that file contains depends on the AMI line (v6.1 adds an OpenSearch entry), so an added or reordered entry would silently return a neighbouring credential rather than failing, bringing a cluster up with the wrong RabbitMQ or redis password. Each entry is now selected by label, with a loud failure listing the labels present if one is missing.

OpenSearch auth is unaffected: it comes from `~/.netrc-os`, which v6.1 still writes and which provisioning already syncs grq → mozart → metrics.

### 5. Stop enabling opensearch-dashboards on mozart

v6.1 narrows `hysdsops`' sudoers from a blanket `/usr/bin/systemctl` to specific verb+unit pairs, so `sudo systemctl enable opensearch-dashboards` fails the deploy outright. The call was never needed: dashboards runs on **grq**, where the AMI enables it (active on 5601 on both images); mozart's httpd proxies `/metrics/` to `grq:5601`, and this repo's own `import_dashboard.sh` waits on `GRQ_FQDN:5601`. Verified on a v6.1 cluster with mozart's own service disabled: `https://<MOZART_FQDN>/metrics/api/status` → HTTP 200.

The removed comment asserted the import proxies to `localhost:5601` and that the operasds AMI bake does not enable the service. Both were wrong — the proxy target is grq, and the AMI enables it on the node that serves it.

### 6. Install GDAL into the conda env before activating the venv

`install_cmr_audit.sh` ran `conda install` after sourcing the venv it had just created, which worked only because of conda's launcher: the v6.1.2 conda env used an absolute interpreter, v6.4.2 uses `#!/usr/bin/env python`, so with a venv active it resolves to the venv's python and dies with `ModuleNotFoundError: No module named 'conda'`. GDAL goes into the conda env either way, so it now runs before activation. Also stops calling `deactivate` unguarded, which with no venv active resolves to conda's non-executable `bin/deactivate` on PATH.

### 7. Stage config uploads instead of deleting them first

`etc/settings.yaml`, `etc/harikiri.yml` and `etc/spot_termination_detector.yml` are bind-mount sources for job containers. Deleting one before re-uploading leaves a window where the path does not exist; a `docker run` landing in it makes the daemon create the mount source as a root-owned directory, failing the job with exit 125 and then wedging the path against every later upload. Each file is now rendered beside the live one and renamed over it.

### 8. Fix the verdi snapshot bake's config uploads

Baking a `Verdi=v6.4.2` snapshot failed at `scp: ${HOME}/verdi/etc: No such file or directory`. Two causes, neither AMI-version related (it reproduced on v5.3 and v5.4.3): the destination is passed to scp without shell expansion, so it looked for a literal `${HOME}` directory; and those parent directories are not on the builder image, which a `file` provisioner cannot create. Uploads now use paths relative to the ops home, with a `remote-exec` creating the directories first.

## Verification

Deployed from this branch on v6.1 AMIs with HySDS v6.4.3, on two clusters running **real PGEs** concurrently. Each venue deploys from its own branch (`OPERA-2557-<venue>`) because `mozart.tf` derives both the container tag and a Jenkins job name from `pcm_branch`, and `remove_job` would otherwise delete a concurrent deploy's job mid-build.

### `dev-e2e-pge-DISP_S1-all_phases` (pyoon) — PASSED

`Apply complete! Resources: 365 added` — rc=0 after 24h13m.

```
SUCCESS: Found 31 expected L3_DISP_S1 datasets            31/31
SUCCESS: Found 54 expected L2_CSLC_S1_COMPRESSED datasets 54/54
SUCCESS: frame 17235 phases: historical_01[15], forward_01[1], historical_02[15], forward_02[2...]
SUCCESS: compressed CSLC boundaries at ['20170906', '20210804'] - 2 k-sets across 2 historical phases
SUCCESS: every historical/no_run phase KSC is superseded
```

Exact counts plus all three phase-structure assertions, so the phased-historical walk and the compressed-CSLC lineage resets both behave as designed on v6.4.3.

### The HC-645 fix, proven on the live cluster

This is the check no previous smoke test performed. Driven through OPERA's real
`es_conn_util` path, against the deployed GRQ, with the interim cap reverted so the
framework's own client is in play:

```
opensearch-py: 3.2.0          <- uncapped, as v6.4.3 ships
server:        3.6.0
engine:        opensearch
docs indexed:  10500
query() returned: 10500
RESULT: PASS - PIT deep pagination works
```

On v6.4.2 this same call raises `AttributeError: 'OpenSearch' object has no attribute 'create_point_in_time'`. Confirmed on the node that the installed `OpenSearchUtility._pit()` carries the fix (`create_pit` present, `create_point_in_time` absent).

### Cluster state

```
hysds            3.3.2
hysds_commons    2.4.1        <- carries HC-645
opensearch-py    3.2.0        <- cap reverted; framework client
elasticsearch    7.13.4
OpenSearch       3.6.0  cluster=common_cluster
lightweight-jobs container-hysds_lightweight-jobs-v2.1.1.sdspkg.tar
process_events   RUNNING, uptime 1 day 16:59 - no restarts
celeryconfig     /run/redis/redis.sock, socket present; probe took its no-op branch
```

Negative evidence, all zero as required: `factotum patched:` 0, `swap_ops_repo` 0, `systemctl enable opensearch-dashboards` 0, `a password is required` 0, `No module named 'conda'` 0, credential-selection failures 0, `create_point_in_time` 0. Container tag for this venue: `container-nasa_opera-sds-pcm:opera-2557-pyoon`.

### `dev-e2e-pge-DISP_S1` (collinss) — historical phases passed, forward still running

| phase | frame | result |
|---|---|---|
| historical | 31241 | **14 L3 / 2 CCSLC** — matches expected |
| phased historical | 24718 | **14 L3 / 3 CCSLC** — matches expected |
| forward | 31241 | in progress (expects 55) |

Final forward counts will be added when it completes.

## Test plan

DISP-S1 smoke on `dev-e2e-pge-DISP_S1` and `dev-e2e-pge-DISP_S1-all_phases`, both with `PGE_SIMULATION_MODE` disabled. Expected counts come from each environment's `datasets_e2e.json`.

- [x] mozart **and** factotum report hysds 3.3.2 / hysds_commons **2.4.1**; no `factotum patched:` line in the provisioning log
- [x] lightweight-jobs registered at **v2.1.1**
- [x] `process_events` RUNNING and not restarting
- [x] `~/.sds/config` carries non-empty `MOZART_RABBIT_*`, `MOZART_REDIS_PASSWORD`, `METRICS_REDIS_PASSWORD`, `OS_USER`/`OS_PASSWORD`
- [x] OPERA's index templates and ISM policies apply cleanly on OpenSearch 3.6
- [x] Phase 1 historical completes with real PGEs (14 L3 + 2 CCSLC, F31241)
- [x] phased historical completes (14 L3 + 3 CCSLC, F24718)
- [x] a GRQ query returning **≥ 10,000 hits** succeeds — the HC-645 PIT path, which no prior smoke test exercised
- [ ] forward: the evaluator chain fires end to end (L2_CSLC_S1 → cycle_evaluator → CSC → k_cycle_evaluator → KSC → SCIFLO_L3_DISP_S1)
- [x] all-phases run reaches its expected counts on F17235 (31 L3 / 54 CCSLC) plus all three phase-structure assertions
- [ ] factotum `user_rules_*` logs: no TypeError/Traceback, per-rule verdicts preserved, settled-probe lines present
- [ ] forced job failure auto-retries to completion
- [ ] a failed job's doc keeps its real error and exit code (HC-639 guard)

## Notes for the reviewer

**This PR modifies none of the smoke-test files.** Not `datasets_e2e.json`, `check_pcm.py` or `run_smoke_test.sh` — the diff is provisioning, `conf/sds/cluster.py`, `install_cmr_audit.sh` and `setup.py`. Forward-count expectations are OPERA-2561's and are authoritative; any `test_forward_expected_datasets` behaviour is inherited from `develop`, not introduced here.

**Behaviour changes to expect, not regressions.** Redis job-status keys now live `max(default, time_limit + 2 × hard_time_limit_gap)`, so long-`time_limit` DISP-S1 jobs hold keys hours longer — worth watching mozart redis memory. Celery errors surface real `WorkerExecutionError` / `OrchestratorExecutionError` instead of `UnpickleableExceptionWrapper`. lightweight-jobs v2.1.1 makes a retry of a missing job fail visibly instead of silently succeeding, and waits up to ~100s on the JobLock when retrying a STARTED job. osaka HTTP downloads of `Content-Encoding: gzip` responses now land decoded. Container docs may carry a `urls` field with `url: null`.

### Ops promotion needs two checks

**1. GRQ snapshot restore across the OpenSearch major.** Verified supported: an OpenSearch 2.9 snapshot of OPERA's real `grq_os_templates` indices restores into 3.6 with documents, `_source`, mappings and aliases intact. But restore refuses more than one major back, based on an index's *original* creation version — so a GRQ index first created under 1.x is blocked no matter how long it has been carried forward. Audit before promoting:

```bash
curl -sk --netrc-file ~/.netrc-os \
  "https://<GRQ>:9200/grq_*/_settings?human=true&filter_path=**.version.created_string" \
  | python3 -m json.tool
```

Anything reporting 1.x needs reindexing first. (`human=true` is required, or `created_string` is absent.)

**2. Snapshot shard blobs land outside `base_path` on OpenSearch 3.x — fix must land with this PR.** This is a consequence of the AMI move that is not visible in this diff, because the fix lives in another repo. On 3.x a repository's shard-level blobs are written to a hashed prefix at the **root of the bucket**:

```
s3://<bucket>/<HASH>/<base_path>/indices/<index-uuid>/<shard>/__*
```

Only repo-root metadata stays under `base_path`. This is OpenSearch's `PathType.HASHED_PREFIX`, not a misconfiguration — our `base_path` is correct and neither our terraform nor the AMI sets anything path-related. The OpenSearch docs are stale here: they say `shard_path_type` defaults to `FIXED`; measured behaviour on 3.6.0 is `HASHED_PREFIX`.

Restores of hashed snapshots work. The damage is operational:

- **`terraform destroy` stops purging.** `modules/common/main.tf::null_resource.destroy_es_snapshots` does a prefix-scoped `aws s3 rm --recursive s3://<es_snapshot_bucket>/<project>-<venue>-<counter>`, which cannot see `<HASH>/opera-<venue>-<counter>/...`. One teardown stranded **38,055 objects / 416 MiB** at the root of `opera-dev-es-bucket`, and every v6.1 venue keeps accreting.
- **Snapshot preservation silently breaks.** The "copy the repo prefix out before destroy" procedure now copies only metadata; all shard data is missed, and the archive looks complete but will not restore.

Fix: [pcm_commons #97](https://github.jpl.nasa.gov/IEMS-SDS/pcm_commons/pull/97) pins `shard_path_type: FIXED`. Verified safe on older engines — OpenSearch and Elasticsearch do not validate repository setting names, so pre-3.x accepts and stores it without acting on it (tested on OpenSearch 1.3/2.9/2.19/3.6 and Elasticsearch 7.17 with the `s3` repo type and the exact settings body the tool sends). **This PR introduces the condition; #97 contains it.**

### Latent trap, deliberately not changed here

The metrics `venv_cmr_audit` resolves elasticsearch 9.5.0 and numpy 2.5.2, because the `[cmr_audit]` extra pins neither and does not pull `hysds_commons` in to constrain them. It is inert today: the three cron-run tools (`cmr_audit_hls`, `cmr_audit_slc`, `cmr_audit_dist_s1`) import no search client at all, and `cmr_audit_disp_s1`, which does reach the chain, cannot run in that venv regardless since `hysds` is not installed there. Worth knowing before someone adds a GRQ query to a cron-run audit tool.


pymonger:
## Deploy + smoke test evidence

Deployed this branch end to end on `opera-collinss-0002` — a clean build on **v6.1 AMIs** with **HySDS v6.4.2**, running **real PGEs** (`dev-e2e-pge-DISP_S1`, `PGE_SIMULATION_MODE` off).

```
Apply complete!   terraform rc=0   2026-08-15T20:36:55Z
```

### Cluster state after deploy

```
hysds 3.3.2 / hysds_commons 2.4.0        framework bump
opensearch 3.6.0, cluster=common_cluster  3-node across mozart/grq/metrics
11 OPERA index templates applied          templates/ISM fine on 3.6
container-hysds_lightweight-jobs-v2.1.0   from the framework bundle
celeryconfig -> /run/redis/redis.sock      socket present (srwxrwxrwx redis redis)
process_events RUNNING, uptime 2 days     never crash-looped
MOZART_RABBIT_USER / MOZART_RABBIT_PASSWORD / MOZART_REDIS_PASSWORD /
METRICS_REDIS_PASSWORD / OS_USER / OS_PASSWORD  -- all non-empty
```

### Per-change evidence

Most of these changes exist so that a failure *stops* happening, so the evidence is largely counts that must be zero:

| change | evidence from the deploy |
|---|---|
| hot-patch retired | `factotum patched:` occurrences **0**; `swap_ops_repo` occurrences **0** |
| dashboards sudo removed | `systemctl enable opensearch-dashboards` attempts **0**; `sudo: a password is required` **0** |
| credentials by label | credential-selection failures **0**; all six values non-empty in `~/.sds/config` |
| conda / venv fix | `No module named 'conda'` occurrences **0**; `install_cmr_audit.sh` completed |
| redis socket probe | celeryconfig and the on-disk socket agree at `/run/redis/redis.sock`; helper took its no-op branch |
| staged config uploads | `.new` + rename observed on both mozart and factotum: |

```
[factotum] put: .../verdi/etc/settings.yaml.new
[factotum] run: mv -f verdi/etc/settings.yaml.new verdi/etc/settings.yaml
[factotum] put: .../verdi/etc/harikiri.yml.new
[factotum] run: mv -f verdi/etc/harikiri.yml.new verdi/etc/harikiri.yml
[factotum] put: .../verdi/etc/spot_termination_detector.yml.new
[factotum] run: mv -f verdi/etc/spot_termination_detector.yml.new verdi/etc/spot_termination_detector.yml
[mozart]   put: .../mozart/etc/settings.yaml.new
[mozart]   run: mv -f mozart/etc/settings.yaml.new mozart/etc/settings.yaml
```

### Smoke test — Phase 1 (historical, real PGEs): PASS

Matches the harness baseline exactly.

```
L3_DISP_S1              14/14
L2_CSLC_S1_COMPRESSED    2/2
check_datasets_file: SUCCESS
```

### Smoke test — Phase 2 (forward): pipeline correct, fixture stale

The trigger chain is exact — every complete non-blackout CSC produced a KSC, zero drops:

```
CSC complete                            56
  of which blackout                     30
  matching the kce rule (non-blackout)  26
KSC total                               26     <- 1:1, gap 0
L3_DISP_S1 total                        25     (14 historical + 11 forward)
```

Supporting evidence that HC-633's machinery is live and healthy: **115** settled-probe entries in the factotum `user_rules_dataset` logs, **0** evaluator errors/tracebacks, 160 `user_rules` workers running, per-rule verdicts logging normally.

`test_forward_expected_datasets` nevertheless **fails**, because `datasets_e2e.json` expects **55** forward `L3_DISP_S1`:

```
fwd L3_DISP_S1 expectation set   2026-03-16  (5d2fbae4)
blackout-in-forward (OPERA-2559) 2026-07-20  (b2019acc, #1458)
```

The expectation predates blackout filtering by four months. It was calibrated when all 56 forward dates could yield products; with blackout active, 30 are excluded by design, leaving 26 eligible and 11 fired. This was never caught because the previous cluster's Phase 2 was deliberately short-circuited, so no post-blackout Phase 2 baseline exists.

**This failure is not specific to this PR** — `test_forward_expected_datasets` will fail on any branch containing OPERA-2559, including `develop`.

Fixing it is a separate decision, and I would value input on which way to go: re-baseline the count (simple, but brittle — it moves whenever the blackout file changes), or make the check blackout-aware and assert the invariant that actually matters, `complete non-blackout CSC == KSC`, which held exactly here.

### Still open

- `elasticsearch==7.13.4` client path against the 3.6 server (`daac_data_subscriber.py`, audit tools via `opera_commons/es_connection.py`) — not yet exercised
- ops promotion still needs the GRQ snapshot-restore check across the OpenSearch major

pymonger:
### Correction on the forward-count check

My earlier comment suggested re-baselining `datasets_e2e.json`'s forward expectation. Disregard that — the counts in OPERA-2561 are the correct ones, and this PR is not the place to change them.

For the record, **this PR modifies none of the smoke-test files** — not `datasets_e2e.json`, `check_pcm.py`, or `run_smoke_test.sh`. Its 13 files are provisioning, `conf/sds/cluster.py`, `install_cmr_audit.sh` and `setup.py`. The `test_forward_expected_datasets` failure observed in the run above is a pre-existing condition on `develop`, not something introduced or owned here.

**Sequencing:** merge OPERA-2561 first, pull its updates into this branch, then re-run the smoke test in that combined state — which is also the state the forward numbers were calibrated for.

What the run above does establish still stands on its own, and none of it depends on the forward count:

- HySDS v6.4.2 installed and healthy (hysds 3.3.2 / hysds_commons 2.4.0)
- OpenSearch 3.6.0 on the v6.1 AMIs, `common_cluster` formed, 11 OPERA index templates applied
- Phase 1 historical with real PGEs: **14 L3 + 2 CCSLC**, matching baseline exactly
- Trigger chain exact: **26 complete non-blackout CSC → 26 KSC**, zero drops (115 settled-probes, 0 evaluator errors)
- `process_events` running 2 days without a restart; all six label-selected credentials populated
- Hot-patch fully retired: 0 `factotum patched:`, 0 `swap_ops_repo`

pymonger:
### Final artifact snapshot (captured immediately before teardown)

Taken from `opera-collinss-0002` at 2026-08-17T13:21Z, just before the cluster was destroyed.

```
--- mozart: framework, redis, process_events ---
hysds 3.3.2  hysds_commons 2.4.0
REDIS_UNIX_DOMAIN_SOCKET = "unix:<redacted>@/run/redis/redis.sock"
srwxrwxrwx. 1 redis redis 0 /run/redis/redis.sock
process_events   RUNNING   uptime 4 days, 0:18:45
container-hysds_lightweight-jobs-v2.1.0.sdspkg.tar

--- OpenSearch ---
opensearch 3.6.0  cluster=common_cluster
100.104.40.171  dimr  -   3.6.0
100.104.40.202  dimr  *   3.6.0
100.104.40.88   dimr  -   3.6.0

--- final tallies ---
CSC complete:           56
CSC complete+blackout:  30
CSC matching kce rule:  26
KSC total:              26      <- 1:1, zero dropped triggers
L3_DISP_S1 v1.0:        25      (14 historical + 11 forward)
L2_CSLC_S1_COMPRESSED:   2

--- factotum: HC-633 evaluator health ---
settled-probe entries: 283
evaluator errors:        0
user_rules RUNNING:    160

--- smoke test ---
SUCCESS: Found 1 expected area_of_interest datasets of version '2.0'
SUCCESS: Found 14 expected L3_DISP_S1 datasets of version 'v1.0'
SUCCESS: Found 2 expected L2_CSLC_S1_COMPRESSED datasets of version '1'
ERROR:   Failed to find 55 expected L3_DISP_S1 datasets of version 'v1.0'   <- pre-existing, see previous comment
forward driver: FIRED 11 / FAIL 30 (blackout) / no-fire 16
```

`process_events` holding a 4-day uptime is the sharpest single result here: that daemon is what a redis socket mismatch would have crash-looped, and it ran the entire deploy and smoke test without a restart.

Full artifact set (11 files, including per-change negative evidence and the deploy chain log) is retained off-cluster at `~/dev/opera/tmp/OPERA-2557-artifacts/`. Cluster torn down; nothing further depends on it.

pymonger:
## Open items closed — and one real defect found

Ran both remaining items against live OpenSearch servers (2.9.0 = v6.0 AMI engine, 3.6.0 = v6.1 AMI engine, security plugin on). One of them turned up a genuine break that the smoke test could not have caught.

### Item 1 — the client path was mis-framed, and it hides a real bug

The item was written as "does `elasticsearch==7.13.4` work against a 3.6 server". That framing is wrong. `conf/settings.yaml` sets `GRQ_ES_ENGINE: opensearch`, and `pcm_commons` dispatches on it:

```python
class PcmUtility(ElasticsearchUtility, OpenSearchUtility):
    def __init__(self, es_url, engine="elasticsearch", ...):
        if engine == "opensearch":
            OpenSearchUtility.__init__(self, es_url, **kwargs)
```

So every GRQ/Mozart call OPERA makes goes through **opensearch-py**, not the elasticsearch client. The elasticsearch pin still matters, but only because `PcmUtility` imports *both* utils modules, so `hysds_commons/elasticsearch_utils.py` executes `from elasticsearch import RequestsHttpConnection` at module scope — a symbol removed in elasticsearch-py 8.x.

**The defect:** `hysds_commons` 2.4.0 `OpenSearchUtility._pit()` calls `create_point_in_time()` / `delete_point_in_time()`. opensearch-py 3.x removed both, renaming them `create_pit`/`delete_pit`. And hysds_commons asks for `opensearch-py>=2.3.0` with **no upper bound** — HC-606 dropped the `<3.0.0` cap as part of the OpenSearch 3.x work in v6.4.2.

The collinss deploy accordingly installed **opensearch-py 3.2.0**:

```
elasticsearch-9.5.0     <- venv_cmr_audit only, see below
numpy-2.5.2
opensearch-py-3.2.0
```

`SearchUtility.query()` switches to `_pit()` whenever the engine is opensearch and the result set reaches **10,000 hits**. Every such query raises `AttributeError`. The smoke test never hit it because no index there held 10,000 documents — at ops scale the subscriber catalogs, the audit tools and `retrieve_disp_s1_from_grq` all cross that line.

Matrix, exercising the real `AncillaryUtility -> PcmUtility -> OpenSearchUtility` chain over 16 operations (index, get, count, search, query, partial update + nested-map merge, msearch, index template, ISM policy, and PIT deep pagination across 10,500 docs):

| client | server | result |
|---|---|---|
| opensearch-py 2.8.0 | 2.9.0 | **16/16 pass** — pre-upgrade baseline |
| opensearch-py 2.8.0 | 3.6.0 | **16/16 pass** — the fix |
| opensearch-py 3.2.0 | 2.9.0 | 15/16, PIT fails |
| opensearch-py 3.2.0 | 3.6.0 | 15/16, PIT fails |

The failure tracks the **client, not the server**. opensearch-py 2.8.0 drives OpenSearch 3.6.0 correctly for everything OPERA does, so capping costs no 3.x server support.

Fixed in `6739c3e9` — `opensearch-py<3.0.0` in `[docker]` and `[subscriber]` (repeated because the mozart venv installs that extra as a separate pip resolution). Verified `pip install .[subscriber]` now resolves 3.2.0 → 2.8.0.

**This is not OPERA-specific.** hysds_commons `develop` still calls the removed methods with an uncapped requirement, so any v6.4.2 deployment is exposed, NISAR included. Worth a HC ticket.

*Side observation, no change made:* `venv_cmr_audit` on metrics resolved elasticsearch 9.5.0 + numpy 2.5.2, because `[cmr_audit]` pins neither and doesn't pull hysds_commons in to constrain them. It's inert today — the three cron-run tools (`cmr_audit_hls`, `cmr_audit_slc`, `cmr_audit_dist_s1`) import no search client at all, and the one that does reach the chain (`cmr_audit_disp_s1`) can't run in that venv regardless, since hysds isn't installed there. Flagging it as a latent trap rather than touching it.

### Item 2 — snapshot restore across the major: supported

Seeded with OPERA's real `grq_os_templates` (14/14 accepted on every version), snapshotted to an fs repo, restored into 3.6.0.

```
2.9.0 -> 3.6.0   RESTORE SUPPORTED
    grq_1_l3_disp_s1              200 docs   open/green
    grq_1_l2_cslc_s1_compressed     0 docs   open/green
    _source intact; mappings preserved exactly (@timestamp date,
    id text+keyword, metadata.frame_id text+keyword, blackout boolean);
    `grq` alias carried over.

1.3.19 -> 3.6.0  RESTORE BLOCKED
    "cannot restore index [grq_1_l3_disp_s1] because it cannot be upgraded"
```

The promotion path itself is fine. The residual risk is any GRQ index whose **original** creation version predates 2.x — restore refuses more than one major back, no matter how long the index has been carried forward. Pre-promotion check:

```bash
curl -sk --netrc-file ~/.netrc-os \
  "https://<GRQ>:9200/grq_*/_settings?human=true&filter_path=**.version.created_string" \
  | python3 -m json.tool
```

Anything reporting 1.x needs reindexing first.

### Scope limits

Single-node containers, not a 3-node `common_cluster`; no shard relocation or S3-backed snapshot repo exercised. Auth was basic-auth over TLS with demo certs — which matches how `OpenSearchUtility` connects (`use_ssl` hardcoded true, `http_auth` from `~/.netrc-os`) but not the AMI's certificate set.

Both harnesses are retained with the artifact set so this is re-runnable.

pymonger:
### Coupled change in pcm_commons — please don't promote this to ops without it

Flagging a consequence of the AMI bump in this PR that isn't visible from this diff, because the fix lives in a different repo.

Moving the cluster AMIs v6.0 → v6.1 takes OpenSearch from 2.9 to 3.6. On 3.x, a snapshot repository's **shard-level** blobs are written to a hashed prefix at the **root of the bucket**, outside `base_path`:

```
s3://<bucket>/<HASH>/<base_path>/indices/<index-uuid>/<shard>/__*
```

Only repo-root metadata stays under `base_path`. This is OpenSearch's `PathType.HASHED_PREFIX`, not an OPERA misconfiguration — our `base_path` is correct, and neither our terraform/conf nor the AMI's `opensearch.yml` sets anything path-related. (Note the OpenSearch docs are stale here: they say `shard_path_type` defaults to `FIXED`. Measured behaviour on 3.6.0 is `HASHED_PREFIX`.)

**Restores of hashed snapshots work fine.** The damage is operational, and there are two parts:

1. **`terraform destroy` stops purging.** `modules/common/main.tf::null_resource.destroy_es_snapshots` does a prefix-scoped `aws s3 rm --recursive s3://<es_snapshot_bucket>/<project>-<venue>-<counter>`, which cannot see `<HASH>/opera-<venue>-<counter>/...`. Tearing down `collinss-0002` on 8/17 left **38,055 objects / 416 MiB** stranded at the root of `opera-dev-es-bucket`. Every v6.1 venue keeps accreting.
2. **Snapshot preservation silently breaks.** The established "copy the snapshot repo prefix out before destroy" procedure now copies only metadata — all shard data is missed. The archive looks complete and will not restore.

Current state of `opera-dev-es-bucket`: **205 hashed prefixes vs 12 real venue prefixes**. Only v6.1 venues produce them; every v6.0/OS-2.9 venue is clean. The same will hit `opera-ops-es-bucket` the moment ops promotes.

**Fix:** [pcm_commons #97](https://github.jpl.nasa.gov/IEMS-SDS/pcm_commons/pull/97) pins `shard_path_type: FIXED` on the repository in `create_snapshot_repository_s3`. Verified safe on older engines — OpenSearch and Elasticsearch don't validate repository setting names, so pre-3.x accepts and stores it without acting on it (tested on OpenSearch 1.3/2.9/2.19/3.6 and Elasticsearch 7.17, with the s3 repo type and the exact settings body the tool sends).

These two should land together. This PR is what introduces the condition; #97 is what contains it.

The `collinss-0002` orphans have already been cleaned up (0 objects remaining).

hhlee445 (approved):

PR #1477, opened 2026-08-18
OPERA-2610: Add DISP-S1 large-gap purge utility

pymonger:
## Summary

Adds a utility that removes CCSLC and DISP-S1 products belonging to k-cycles that span a sensing-time gap longer than two years, and produces the removal list ASF DAAC needs for their copies.

Frames that stopped acquiring — most of them across the S1B outage — and resumed years later were processed on the absolute k=15 grid, so the ministack straddling the gap phase-linked pre-gap and post-gap acquisitions. That is not a valid interferometric stack, and every compressed CSLC after it inherited the problem through the lineage. The large-gap guard that now refuses these stacks was merged after the historical catch-up batch procs ran and is not deployed on OPS-POP1, so the products are still in the system and were delivered to the DAAC. This purge is the pre-clean for the gap-aware (phased) historical reprocessing in OPERA-2561.

Resolves [OPERA-2610](https://hysds-core.atlassian.net/browse/OPERA-2610)

## Changes

New files only — nothing existing is modified.

**`tools/delete_disp_s1_large_gap.py`** — the utility, with four subcommands:

- `audit` — inventories GRQ, S3 and CMR independently, derives each frame's gap and boundary, and writes the deletion manifests plus a summary. Read-only.
- `asf-list` — turns an audit into a CSV removal list for the DAAC.
- `execute` — deletes the manifest rows from GRQ and S3.
- `verify` — re-inventories and checks the affected sets are gone and the kept sets are untouched.

How the affected set is chosen, per frame:

1. `gap_start_date` — the last sensing date before the frame's first gap longer than `--gap-days` (default 730), from the consistent burst database; from the `--gap-list` report when the frame is not in the database.
2. `last_clean_boundary_date` — the most recent CCSLC `last_date` at or before the gap start. This is anchored on the CCSLCs that actually exist rather than on burst-database positions, because processing ran against an older database vintage than the one used for analysis. It falls back to burst-database k-cycle math when no CCSLCs are present, and a frame with fewer than k pre-gap dates has no clean boundary at all.
3. Affected CCSLCs — `last_date` after the gap start.
4. Affected DISP-S1 products — secondary date after the boundary; every product when there is no boundary.

Everything at or before the boundary is valid history and is kept.

Safety model:

- Dry-run is the default for every subcommand; `execute` mutates nothing without `--execute`, and then only after a typed `DELETE` (`--yes` for scripted runs).
- Selection happens once, in `audit`. `execute` consumes the reviewed manifests and never recomputes what to delete, so what a reviewer approves is exactly what is removed.
- Before any delete, an NDJSON backup of every document is written and the documents are copied into `parked_disp_s1_large_gap_*` indices.
- `--max` caps the blast radius of a single run (default 10,000 items, counting GRQ docs plus S3 datasets); S3 prefixes shorter than a dataset directory are refused; deletes are idempotent.

Products are located by parsing granule ids, so a product that aged out of GRQ but still sits in S3 (or the reverse) is still found. Most affected DISP-S1 products no longer exist locally at all, which is why the removal list is built from CMR rather than from GRQ.

Staged `L2_CSLC` metadata is deliberately left alone: a burst belongs to several frames, so deleting those documents by frame could damage a frame that is not in scope. Cycle-state-configs are frame-scoped and are handled behind an opt-in `--state` flag.

**`tests/unit/test_delete_disp_s1_large_gap.py`** — 58 tests over the burst-database gap math, granule-id parsing, boundary derivation, classification, GRQ/S3 reconciliation, the manifest contract, and the safety rails. Fixtures mirror frame 24726's real shape on OPS-POP1 (86 pre-gap dates, a 1259-day gap, five clean CCSLC boundaries plus one spanning the gap). The end-to-end cases drive the CLI against fake OpenSearch, S3 and CMR clients and assert that dry runs make no mutating calls at all.

**`tools/DELETE_DISP_S1_LARGE_GAP_README.md`** — selection rules, safety model, invocation, outputs and the anomalies the audit reports.

## Documentation

Operator procedure, with the script attached: [DISP-S1 Large-Gap Frame Purge (CCSLC, DISP-S1, and DAAC removal list)](https://wiki.jpl.nasa.gov/spaces/operasds/pages/1310026272/DISP-S1+Large-Gap+Frame+Purge+CCSLC+DISP-S1+and+DAAC+removal+list) under the operasds Procedures space.

## Test plan

- [x] 58 unit tests pass
- [x] Audited all six priority lists on OPS-POP1 (335 frames) against the 2026-08-10 consistent burst database
- [x] Rehearsed every `execute` phase in dry-run and generated all six removal lists
- [x] Zero-mutation proof: cluster document counts identical before and after every dry run (CCSLC 309,637 and DISP-S1 14,150), and no parked indices created
- [x] Ground truth confirmed on six frames against independently derived expectations: 24726 (27 CCSLCs affected of 297, 15 products of 89, boundary 20210806), 44325 (27 affected, boundary 20211127), 5127 (already clean — its pre-gap series is exactly k-aligned), 11109 and 26432 (no affected products; their gaps only appear in a burst database newer than the one processing used), 26425 (nothing present)
- [x] Selected granules inspected by hand: affected CCSLCs are exactly the gap-spanning window, and affected products start one cycle past the boundary
- [ ] Production execution on OPS-POP1 — pending; to be run from the wiki procedure with a review gate on the manifests and a single-frame canary before the priority runs

Audit result over the 335 frames: 45 have something to purge and 290 were never processed past their gap; 3,443 CCSLCs and 2,461 DISP-S1 products are affected, of which only 405 still exist locally, and 28,997 CCSLCs and 18,720 products are kept.


hhlee445 (approved):
LGTM

PR #1479, opened 2026-08-18
Fix OOM crash in daac_data_subscriber survey mode for large date ranges

kevinmarlis:
## Purpose
- The survey subcommand crashes with out-of-memory errors when querying large date ranges (e.g., 10 years of CSLC-S1 data) because all CMR results are accumulated in a dict before processing begins.
- NOTE: this does not make use of the same CMR cache to disk memory fix as applied to cmr_audit scripts. The nature of the survey script allows CMR results to be processed and discarded on the fly without a major refactor, unlike the other scripts.

## Proposed Changes
- [CHANGE] Process each time window's CMR results immediately as futures complete, then free the granule list, instead of storing all windows' results simultaneously in memory.

## Issues
- OPERA-2615

## Testing
- Logic is identical to before — only the timing of when results are consumed changed. Output CSV files and histogram are produced in the same order with the same content.

### DIFF
--- a/data_subscriber/survey.py
+++ b/data_subscriber/survey.py
@@ -69,7 +69,6 @@ def run_survey(args, token, cmr, settings):
 
         start_dt = start_dt + step_time
 
-    cmr_results = {}
     concurrency = 5
     logger.info(f"Preparing to issue concurrent CMR queries. {concurrency=}")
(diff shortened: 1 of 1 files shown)


hhlee445 (approved):
LGTM

PR #1485, opened 2026-08-25
Patch: Fixes & improvements for DISP-S1 FWD driver script

RKuttruff:
## Purpose
- Ensure KSC `region_id` is fetched in ES query. Fixes KSCs always being marked as non-firing due to being seen as in region="UNKNOWN", which is not whitelisted in the DISP ST pipeline
- Check CSC blackout status to short circuit the wait for KSCs in a blackout window

## Testing
- [x] DISP-S1 ST pipeline


pymonger (approved):
lgtm

PR #1488, opened 2026-08-26
OPERA-2477: Fix snapshot schedules and retain product data indefinitely

pymonger:
## Summary

OPERA clusters snapshot their OpenSearch indices to S3 on a schedule, driven by Snapshot Management policies that provisioning installs on each cluster. Three things were wrong with those policies: the schedule did not mean what it says, the retention window was far shorter than intended, and product data shared a policy with operational data, so the two could not be retained differently.

This fixes all three, guards the archive at teardown, and adds a runbook for getting data back out.

Resolves [OPERA-2477](https://hysds-core.atlassian.net/browse/OPERA-2477).

> **Depends on [IEMS-SDS/pcm_commons#98](https://github.jpl.nasa.gov/IEMS-SDS/pcm_commons/pull/98), which must merge first.** Provisioning here passes `--no-deletion`, an option that PR adds. Deploying this branch against an older `pcm_commons` fails on an unrecognised argument.

## What was wrong

**Snapshots on the three-cluster layout ran monthly, not daily.** All three policies passed `--schedule="0 0 5 * * ?"`, a six-field expression meaning 5 AM daily. Snapshot Management keeps only the leading five fields, so it was stored as `0 0 5 * *` — midnight UTC on the fifth of the month.

**Retention is about three days on both production clusters today.** Every policy carried `max_age 60d / min_count 10 / max_count 60`. On an hourly policy the count is what binds: 60 hourly snapshots is two and a half days. Measured on the live clusters, both hold snapshots spanning only **3 days 6 hours**, and a single deletion pass on ops-fwd trimmed 24 of them.

**Product and operational data shared one policy.** On a combined cluster, catalogs and products sit in the same repository as job, worker and log indices, so there was no way to keep the first indefinitely while letting the second roll off — which is the requirement.

**Index selection was an allowlist.** `grq_*,*_catalog` misses date-rolled catalogs such as `hls_catalog-2026.08` outright, and misses any index family added after the list was written.

## What this changes

Index selection is now expressed by **exclusion** rather than by listing, so a family introduced later is captured by default instead of silently missed. System indices stay out, and `jobs_accountability_catalog*` stays out deliberately — the project does not preserve it, and it is among the largest things in the archive.

On a combined cluster the single policy becomes two over the same repository, staggered half an hour apart so they never start together:

| policy | schedule (UTC) | holds | retention |
|---|---|---|---|
| `hourly-snapshot-grq` | `0 * * * *` | products and catalogs | kept until deleted deliberately |
| `hourly-snapshot-mozart-metrics` | `30 * * * *` | job, worker and log indices | 60 days, unchanged |

On the three-cluster layout, all three schedules become five-field `0 5 * * *`, and the GRQ policy takes the same exclusion selection with no deletion block. Reverting to a trimmed archive later is a single policy update, documented in the runbook.

**No policy sets a snapshot time limit.** The shared tool hardcoded one hour on both halves of every policy. OpenSearch applies no limit of its own, and exceeding one does not stop a snapshot — Snapshot Management stops tracking it and marks the run failed while the snapshot carries on in the cluster. A limit shorter than a snapshot really takes therefore reports failures for work that is succeeding, and frees the next scheduled run to start a second snapshot over the first. It is not biting on either production cluster today (the slowest run in the retained window was 9.6 minutes on fwd, 13 on pop1), but the first snapshot of a new repository is a complete copy and runs for hours.

At teardown, the branch that deletes a venue's archive now refuses to run for a venue named `ops*`, on top of the per-venue default, since that archive cannot be reproduced.

## Measured impact on the production clusters

Both selections were resolved directly against ops-fwd and ops-pop1, read-only — asking each cluster what its own indices resolve to, rather than matching names offline:

| cluster | selected before | selected after | products / operational | no longer selected | newly selected |
|---|---|---|---|---|---|
| ops-fwd | 367 | 365 | 46 / 319 | 2× `jobs_accountability_catalog-*` | none |
| ops-pop1 | 213 | 211 | 59 / 152 | 3× `jobs_accountability_catalog-*` | `parked_ccslc_orphans_20260623` |

**Nothing stops being captured except the deliberate exclusion**, and the two policies divide the namespace with no index captured twice.

The immediate coverage gain is therefore small — one index, on pop1. The value of the change is the retention split and the schedule fix, plus the fact that the next index family will be captured without anyone editing a pattern. Worth noting for anyone reading the ticket first: it names `accountability` and `duplicates` as silently missed, and neither exists on either cluster today.

## Changes

**`cluster_provisioning/modules/common/mozart.tf`** — three-cluster GRQ, Mozart and Metrics policies move to five-field `0 5 * * *`; the GRQ one takes the exclusion selection and `--no-deletion`. The combined-cluster policy becomes the two policies above. Selections are quoted, since they contain `*` and leading `-` and are interpolated into a remote shell. The comment block explains the schedule truncation, the retention split, and why selection is by exclusion.

**`cluster_provisioning/modules/common/main.tf`** — the teardown provisioner refuses to purge for an `ops*` venue; it removes all three policy names so a cluster built either way tears down cleanly, tolerating a missing one rather than aborting a destroy; and the final teardown snapshot uses the same selection as the policies rather than the stale allowlist.

**`cluster_provisioning/ES_SNAPSHOTS.md`** (new) — what is preserved and why, restoring selected indices or a whole cluster, re-enabling rolling deletion, retrofitting a running cluster, bucket hardening, and the index-age audit an OpenSearch major upgrade needs. Two warnings there matter even if the rest is skipped: never expire snapshot objects with an S3 lifecycle rule, because snapshots share segment files and deleting them from the S3 side corrupts every snapshot that referenced them; and never transition them to Glacier.

## Testing

**Terraform.** Formatting and validation output are byte-identical to develop: the pre-existing differences are unchanged and none are introduced.

**Both OpenSearch versions.** Containers running 2.9.0, which production runs today, and 3.6.0, which development runs and which production moves to on the next AMI line. Every policy-creating command is taken from `mozart.tf` and run verbatim, then each installed policy is read back and a real snapshot is taken with that policy's own selection to see what it captures. 102 checks, all passing. This also reproduces the schedule defect on a real Snapshot Management: the old six-field expression really is stored as monthly.

**A live cluster, end to end.** `gmanipon-0036` on OpenSearch 3.6.0, deployed from this branch together with its companion: 381 resources added, no errors. Both policies installed with the schedules, absent time limits and retention shown above, and fired six times, all successful.

To test selection against something realistic, the **combined index namespace of ops-fwd and ops-pop1 was replicated onto the cluster — 456 index names — along with 44 real documents sampled from both**. With that namespace present, the products policy captured 92 indices: no system indices, no `jobs_accountability_catalog`, no job or log indices, and no overlap with the other policy. Snapshotting with the old and new selections and comparing them loses exactly the three `jobs_accountability_catalog-*` indices and nothing else.

**Teardown and recovery.** Destroying the cluster with the archive-preserving setting left the archive byte-identical — 6,116 objects, 6,098 shard files — with no instances remaining. Redeploying the same venue re-registered the repository against that archive and listed all eight prior snapshots. Because listing only proves the metadata survived, an index was then restored from a snapshot taken *before* the destroy: it came back with its original document count and real content, on a cluster that did not exist when those documents were written.

**Storage layout.** OpenSearch 3.6 can scatter shard files to a hashed prefix at the root of the bucket instead of under the repository path, which defeats prefix-scoped cleanup and archival. The repository here carries the setting that prevents it, and the layout confirms it: 6,696 shard files under the venue path and no hashed prefix, including after the redeploy re-registered the repository.

## Follow-on, not in this PR

Bucket hardening (versioning, plus a bounded expiry of non-current versions on `opera-ops-es-bucket`) and retrofitting the two running production clusters onto these policies are operational steps rather than code; both are written up in the runbook. The buckets are not managed by this repository's Terraform.

Separately, redeploying into a previously used venue fails a few `aws_cloudwatch_log_group` creations with `ResourceAlreadyExistsException`, because the CloudWatch agent on the new instances recreates log groups while Terraform is also creating them. That is a pre-existing race in `modules/common/autoscaling_groups.tf`, unrelated to this change, and it occurs after provisioning has finished. It deserves its own ticket if redeploying into a used venue is meant to be routine.


hhlee445 (approved):
LGTM

PR #1482, opened 2026-08-19
DISP-S1 state config region lookup and whitelisting in DISP-S1 trigger

RKuttruff:
## Purpose
- Mark CSCs and KSCs with the region associated with their frames
  - Looked up through an ancillary db defined in settings
- Helper tool for operators to set/disable region whitelisting at the disp-s1 trigger level
- [WIP] Testing of whitelisting in the DISP e2e pipeline

This PR is dependent on #1475 to merge

## Issues
- [OPERA-2617](https://hysds-core.atlassian.net/browse/OPERA-2617)

## Testing
- [x] Cluster deploy
- [x] Verified CSC/KSC region metadata
- [x] Tested whitelist config script

Normal DISP trigger:

image

Setting a whitelist:

```
$ python ~/mozart/ops/opera-pcm/tools/disp_s1_set_whitelist.py --whitelist-regions 4
[2026-08-20 21:23:06,330: INFO/get_logger] Added logging filter for elasticsearch_utils/opensearch_utils
[2026-08-20 21:23:06,330: INFO/get_logger] Initial logging configuration complete
[2026-08-20 21:23:06,331: INFO/get_logger] Log level set to INFO
[2026-08-20 21:23:06,332: INFO/get_es_connection] aws_es=False
[2026-08-20 21:23:06,332: INFO/get_es_connection] es_url=['https://100.104.40.170:9200', 'https://100.104.40.235:9200', 'https://100.104.40.99:9200']
[2026-08-20 21:23:06,332: INFO/get_es_connection] es_engine='opensearch'
[2026-08-20 21:23:06,346: INFO/log_request_success] HEAD https://100.104.40.170:9200/ [status:200 request:0.013s]
[2026-08-20 21:23:06,360: INFO/log_request_success] POST https://100.104.40.235:9200/user_rules-grq/_search [status:200 request:0.014s]
[2026-08-20 21:23:06,389: INFO/log_request_success] POST https://100.104.40.99:9200/user_rules-grq/_update/7tnfG6ABOmuiSiy6FUD3?refresh=true [status:200 request:0.028s]
```

Updated trigger:

image

Disabling the whitelist:

```
$ python ~/mozart/ops/opera-pcm/tools/disp_s1_set_whitelist.py --disable-whitelist
[2026-08-20 21:21:41,064: INFO/get_logger] Added logging filter for elasticsearch_utils/opensearch_utils
[2026-08-20 21:21:41,064: INFO/get_logger] Initial logging configuration complete
[...]

PR #1489, opened 2026-08-27
OPERA-2279: Handle DISP-S1 frames with no sensing datetimes in the consistent burst database

pymonger:
## Summary

DISP-S1 assumes every frame in the consistent burst database has at least one sensing datetime. Under the new CONOPS a frame can have a burst pattern and no sensing datetimes. Today the first CSLC for such a frame raises `IndexError` before any decision is made.

This makes those frames process forward, and rejects them where forward processing is not the right answer.

Resolves [OPERA-2279](https://hysds-core.atlassian.net/browse/OPERA-2279).

## What breaks today

Anything that computes an acquisition cycle needs two things the database supplies as a sensing datetime list: an epoch to count from, and a series to locate the acquisition in. A frame with an empty list has neither.

- `sensing_time_day_index` reads `sensing_datetimes[0]` unguarded, so it raises immediately.
- `determine_k_cycle` locates the acquisition with `list.index()` into the same empty list.
- `get_prev_day_indices` slices it to build the compressed CSLC lineage.

Each needs a different answer, not the same guard.

## Approach

**Epoch comes from the database.** A frame with no sensing datetimes has no first acquisition to anchor on, so cycles count from the survey start parsed out of `metadata.input_cmr_csv` and stamped on each frame at load. An acquisition cycle is only a label every component agrees on, so tying it to the database in force keeps it from drifting from a constant written somewhere else. Both deployed databases report 2016-07-01, which is also the fallback if the metadata has no parseable range.

**Series comes from state we already have, not CMR.** K-cycle position counts the cycle state configs the k-cycle evaluator publishes, which exist exactly when a cycle's bursts are all present. The evaluator decides window position and CCSLC boundaries from the same records, so the download side and the evaluator agree by construction. Blacked-out and database-excluded acquisitions are filtered out as everywhere else. This avoids querying CMR to rebuild the series: the answer is already in GRQ.

**Historical is rejected, not guessed.** A frame with no recorded series has nothing to walk historically, so `validate_phased_batch_proc` and `validate_unphased_batch_proc` refuse it with a reason instead of building an empty batch.

## Changes

`data_subscriber/cslc_utils.py` - `_survey_start()` parses the survey range; `_HistBursts` carries `campaign_start`; `frame_sensing_epoch()` returns the frame's own first sensing datetime or falls back to that epoch; `_campaign_day_index()` counts whole calendar days from it. `sensing_time_day_index` routes to it only for a frame that has a burst pattern and no sensing datetimes.

The multiple-of-six rounding that anchors a listed frame is not applied here. There is no acquisition series to round against and applying it would invent structure that is not in the data. Successive acquisitions are days apart, so plain date arithmetic keeps indices distinct and ordered, which is all a cycle label has to be.

`data_subscriber/cslc/cslc_dependency.py` - `_known_complete_cycles()` reads the frame's complete cycle state configs from GRQ, cached per frame, bounded at 1000 with a warning at the ceiling. `get_prev_day_indices` and `determine_k_cycle` take their empty-frame path from it.

`data_subscriber/cslc/cslc_query.py` - the CCSLC search and the reprocessing window are floored at the epoch instead of a sensing datetime that does not exist.

`tools/disp_s1_burst_db_tool.py`, `tools/ops/cmr_audit/diagnose_disp_s1_frame_products.py`, `data_subscriber/asf_cslc_download.py`, `data_subscriber/query.py` - call sites moved to the shared epoch helper.

Note on `frame_to_bursts`: it is a `defaultdict`, so looking up a frame that is not in the database manufactures an empty one that looks identical to a real empty-sensing-datetimes frame. The new path requires a non-empty burst pattern first. A frame we do not process has to keep failing loudly.

## Testing

43 new unit tests, all passing: 17 on the epoch and day-index helpers, 17 on the dependency paths, 9 on the query paths. A fixture database carries both an empty-sensing-datetimes frame and a normal one, so each test asserts the new path is taken and the existing path is unchanged. `test_disp_s1_k_cycle_evaluator.py` gains a check that the download side and the evaluator agree on CCSLC boundaries for such a frame, since that equivalence is the basis for counting cycle state configs.

Full suite against develop, same venv and invocation:

| | develop | branch |
|---|---|---|
| passed | 577 | 626 |
| failed | 1 | 1 |
| collection errors | 5 | 4 |

Same single pre-existing failure on both (`test_complete_workflow_with_real_cmr_data`, which reaches real CMR). No new failures, one collection error fixed.

Two test-infrastructure fixes are included because this branch causes them, each in its own commit:

- `osgeo` ships on cluster images only and `geo.geo_util` imports it, so the query classes these tests cover are uncollectable off-cluster. The stub is in `tests/unit/conftest.py`, not a test module, because injecting it from a test module makes collection depend on import order.
- That stub makes `tests/unit/data_subscriber/test_gcov_query.py` collectable, which exposed a latent bug in it: `test_query_cmr_mocked` rebound `data_subscriber.cmr.async_cmr_posts` and a `gcov_utils` path constant with no cleanup. Both survived the test, so anything running afterwards that queried CMR through that module got the GCOV canned response. Enough to fail the DIST-S1 input tool tests. Both are scoped with `monkeypatch` now.

A third commit makes both DISP-S1 provisioners honour `run_smoke_test`, which they declared and never read. `dev-e2e` already gates on it. The built-in smoke drives 56 forward dates and correctly refuses the historical phase on a stripped frame, so the flag was the control we needed and it was inert. It still defaults to true and nothing in the repo sets it, so no existing caller changes.

## Cluster validation

Three clusters, all deployed from this branch. All destroyed after.

**Empty frame, end to end.** Frames 31241 and 26432 stripped to `sensing_time_list: {}`, matching how the seven real empty frames are encoded. Frame 31241 driven through 17 acquisitions:

```
KSC total=15  incomplete=14  complete=1
BOUNDARY cycle=393  final=True  reason='ready: 15 CSLCs + no CCSLCs required (early window)'
products: L3=1  CCSLC=27, both frame_id=31241 acquisition_cycle=393
jobs: 99 completed, 0 failed
```

Cycles count from the campaign epoch (111, 135, 159 ... 393) and match the offline computation date for date. The k=15 window closes on exactly the 15th acquisition. That frame has a 7-month gap between its first and second acquisitions and a 6-month gap before its seventeenth, and the count is undisturbed by both, because position counts observed complete cycles rather than index spacing.

The CCSLC lineage round-trips. Both acquisitions after the boundary find the CCSLCs it produced, by the same campaign-epoch key:

```
cycle 393  complete=True  save_cc=True   ccslc_ids=0   (early window, none required)
cycle 405  complete=True  save_cc=False  ccslc_ids=27
cycle 417  complete=True  save_cc=False  ccslc_ids=27
```

This is the assertion that matters, since it is where a disagreement between the download side's epoch and the evaluator's would show up.

That venue runs `PGE_SIMULATION_MODE`, so SCIFLO emits fixture output: the CCSLCs carry T041 burst ids and the L3 granule id names a different frame, where 31241 has two T117 bursts. Burst identities and counts are not offered as evidence. What is real is everything upstream of the PGE: cycle numbering, frame attribution, window arithmetic, and the cascade reaching the PGE.

**Phased and gap regression**, frame 17235 on the unmodified database, historical -> forward -> 2.8-year gap -> historical -> forward:

```
L3 31/31   CCSLC 54/54   (datasets_e2e.json)
phases historical_01[15], forward_01[1], historical_02[15], forward_02[2], no_run[9]
compressed CSLC boundaries at ['20170906', '20210804']
9 no_run dates produced no products
every historical/no_run phase KSC superseded
```

**Plain frame regression**, frame 31241 on the unmodified database, anchored at its own first acquisition (0, 210, 222, 234 ...) against 111, 135, 159 ... for the same frame stripped. Dated frames are unchanged. 39 L3, 5 CCSLC, 11 SCIFLO jobs, no failures.

## Issues found that are not from this change

**The all-phases smoke never sets the region whitelist.** `trigger-SCIFLO_L3_DISP_S1` gates on `metadata.region_id` against a whitelist defaulting to `["0"]`, and the test frames are region 4. Counting rule matches with and without that clause isolates it:

```
without region clause -> 3   (exactly the 3 forward dates reported missing)
with    region clause -> 0
```

The base DISP-S1 smoke calls `disp_s1_set_whitelist.py` four times. The all-phases smoke calls it zero times. Region gating came in with OPERA-2617, already in this branch's base, so all-phases forward assertions fail on develop for the same reason. Needs its own ticket.

**`run_disp_s1_forward_serial.py` cannot see products on a simulation-mode cluster.** Line 58 hardcodes `L3_INDEX = "grq_v1.0_l3_disp_s1*"` and simulation-mode PGEs publish `v0.1`, so the driver waits out its L3 timeout and reports failures for products that exist. Same hardcoded pattern is in `capture.py`.

**Re-ingesting a dataset across a monthly index rollover leaves a stale duplicate.** The same `_id` lands in the new month's index while the old copy stays, and wildcard queries see both. A stale `is_complete: false` copy held a k-window at 14/15 until it was removed.

**`tools/reevaluate_disp_s1_forward_kscs.py` refuses a frame with no sensing datetimes** ("frame 31241 has no phases"). Consistent, since no sensing datetimes means no phase labels, but this recovery tool cannot service the frames this ticket introduces.

## Not covered

Frame 26432 through the production `daac_data_subscriber.py query` entry point. Real-PGE product content, by design.

Worth noting from the ticket: it has not been decided at program level whether DISP-S1 forward processing runs in operations. If it does not, the better fix is for ADT to drop sensing-datetime-less frames from the database and this becomes unnecessary.


pymonger:
## Cluster validation — interim report

Three dev clusters, all deployed from this branch. Summary first: **the change behaves correctly on every axis tested**, and the two smoke assertion failures seen along the way are both traced to a region-whitelist gate that predates this branch, with the evidence below.

| cluster | role | burst DB |
|---|---|---|
| `pyoon-0047` | empty-frame forward (the new path) | frames 31241 + 26432 stripped to `sensing_time_list: {}` |
| `gmanipon-0037` | plain-frame regression | unmodified |
| `collinss-0005` | phased/gap regression, frame 17235 | unmodified |

The stripped database matches how the 7 real empty frames are encoded — `{}` with a non-empty `burst_id_list` — rather than the `[]` first assumed.

### The epoch behaves differently for the two frame kinds, on one cluster, from the same granules

`pyoon-0047`, where frame 31241 is stripped and 31240 is intact. They share burst `t117_249921_iw3`, so single granules mint records for both:

```
frame 31241 (stripped) cycles: 111, 135, 159, 183, 207, 231, 243, 255, 321, 333, 345, 357, 369, 381, 393, ...
frame 31240 (intact)   cycles:   0,  24,  48,  72,  96, 120, 132, 144, 210, 222, 234, 246, 258, 270, 282, ...
```

31241 counts from the campaign start (2016-07-01: 2016-10-20 is day 111, 2017-05-18 is day 321); 31240 counts from its own first acquisition. Both values match the offline computation date-for-date.

The same frame on the unmodified database, `gmanipon-0037`:

```
frame 31241 (intact) cycles: 0, 210, 222, 234, 246, 258, 270, 282, 294, 306, 318, 330, 342, 354, ...
```

Cycle 0 is its own first sensing datetime. **A dated frame is unaffected; only the sensing-datetime-less frame changes.**

### The k-cycle boundary lands exactly where it should

`pyoon-0047`, frame 31241, k=15 — 28 KSCs, split 14/14:

```
incomplete: 111, 135, 159, 183, 207, 231, 243, 255, 321, 333, 345, 357, 369, 381
complete  : 393, 405, 417, 429, 441, 453, 465, 477, 489, 501, 513, 525, 537, 549
BOUNDARY cycle=393 final=True reason='ready: 15 CSLCs + no CCSLCs required (early window)'
```

The first 14 cannot be complete (window < k); the 15th is, and carries `save_compressed_cslc`. Position is counted from observed complete cycles, not index arithmetic — which this frame tests hard, since its real history contains a 7-month gap between acquisitions 1 and 2 and a 6-month gap before acquisition 17. The count is undisturbed by both.

### Job outcomes

| | pyoon-0047 | gmanipon-0037 | collinss-0005 |
|---|---|---|---|
| completed | 112 | 354 | 1043 |
| failed | 1 (see note) | 0 | 0 |
| `disp_s1_cycle_evaluator` | 56 | 114 | 891 |
| `disp_s1_k_cycle_evaluator` | 33 | 36 | 94 |

The single failure is a job **I** submitted by hand with malformed parameters while trying to re-trigger a state config; its tag identifies it. The evaluator path under test is 28 completed + 4 deduped on that cluster, no failures. No `IndexError` anywhere — the crash this ticket describes does not occur.

### Phase structure is preserved on a gapped frame

`collinss-0005`, frame 17235 — historical → forward → 2.8-year gap → historical → forward:

```
SUCCESS: frame 17235 phases: historical_01[15], forward_01[1], historical_02[15], forward_02[2], no_run[9]
SUCCESS: compressed CSLC boundaries at ['20170906', '20210804'] (2 k-set(s) across 2 historical phase(s))
SUCCESS: 9 no_run date(s) produced no products
SUCCESS: every historical/no_run phase KSC is superseded
```

Its cycle list shows the gaps intact — `0, 12 … 60`, then `330 … 438`, then `1446+` — anchored at its own first acquisition. CCSLC count is exactly the expected **54/54**.

### The two failures, and why they are not this change

`collinss-0005` reported `ERROR: 3 forward date(s) produced no product: ['20170918', '20210828', '20210909']`, giving 28 L3 against the expected 31, and `pyoon-0047` produced no L3 or CCSLC at all. Same cause in both: the `trigger-SCIFLO_L3_DISP_S1` rule gates on `metadata.region_id` against a whitelist that defaults to `["0"]`, and both test frames are region **4**. The state configs are complete, final and ready; the rule simply cannot match them. Counting rule matches with and without that one clause isolates it:

```
collinss-0005:   rule WITHOUT region clause -> 3      <- exactly the 3 missing forward dates
                 rule WITH    region clause -> 0
pyoon-0047:      rule WITHOUT region clause -> 1      <- the boundary KSC
                 rule WITH    region clause -> 0
```

Region gating came in with OPERA-2617 (#1482, #1485), already present in this branch's base. The base DISP-S1 smoke was updated to set the whitelist; the all-phases smoke was not:

| smoke script | `disp_s1_set_whitelist` calls |
|---|---|
| `dev-e2e-pge-DISP_S1` | 4 |
| `dev-e2e-pge-DISP_S1-all_phases` | **0** |

So the all-phases forward assertions fail on `develop` for the same reason, independent of this branch — which touches only `main.tf` in that directory, not the smoke script and not `datasets_e2e.json`. **Worth its own ticket:** the all-phases smoke needs the whitelist call the base smoke already has.

After setting the whitelist to region 4 on `collinss-0005`, the rule matches 3, and re-driving those dates re-submitted all three with complete burst sets:

```
frame 17235: 3 forward date(s) to re-evaluate
   20170918  csc is_complete=True  found=27/27  blackout=False
   20210828  csc is_complete=True  found=27/27  blackout=False
   20210909  csc is_complete=True  found=27/27  blackout=False
submitted 3/3
```

### Incidental finding

`tools/reevaluate_disp_s1_forward_kscs.py` refuses a sensing-datetime-less frame outright — *"frame 31241 has no phases"*. That is consistent (no sensing datetimes means no phase labels), but it means this recovery tool cannot service the frames this ticket introduces. Not addressed here; noting it for whoever owns that tool.

### Still open

Product generation on the empty-frame path itself is not yet demonstrated end to end — the whitelist blocked it, and the run is being resumed. Everything upstream of the trigger is verified above. The frame 26432 production-entry-point check via `daac_data_subscriber.py` is also still to run. I will post results rather than leave this comment as the final word.

pymonger:
## Cluster validation — results

Follow-up to the interim report above, which said product generation on the empty-frame path was not yet demonstrated. It now is.

### Scenario A — a frame with no sensing datetimes, end to end

`pyoon-0047`, deployed from this branch pointing at a burst database with frames 31241 and 26432 stripped to `sensing_time_list: {}` — the same encoding the seven real empty frames use. Frame 31241 driven through 17 acquisitions.

```
KSC total=15   incomplete=14   complete=1
BOUNDARY cycle=393  final=True  region=4
          reason='ready: 15 CSLCs + no CCSLCs required (early window)'

products:  L3 = 1        CCSLC = 27
           both tagged frame_id=31241, acquisition_cycle=393
jobs:      91 completed, 0 failed
```

The chain that matters: acquisition cycles counted from the campaign epoch (`111, 135, 159 … 393`, matching offline computation date-for-date), the k=15 window closing on exactly the 15th acquisition, the boundary reaching `is_complete` and `compressed_cslc_final` with an empty pending list, the trigger rule firing, and the campaign-epoch cycle **393** appearing in the published product's metadata and in each CCSLC's `ccslc_m_index` (`t041_086865_iw3_393`).

That frame's real history contains a 7-month gap between its first and second acquisitions and a 6-month gap before its seventeenth. The cycle count is undisturbed by both, because position is counted from observed complete cycles rather than index arithmetic.

**The compressed-CSLC lineage round-trip closes.** The acquisition after the boundary locates the compressed CSLCs the boundary produced, by the same campaign-epoch key:

```
cycle 393   complete=True  save_cc=True   ccslc_ids=0  pending=0
      reason: ready: 15 CSLCs + no CCSLCs required (early window)
cycle 405   complete=True  save_cc=False  ccslc_ids=27 pending=1
      uses: OPERA_L2_COMPRESSED-CSLC-S1_..._T041-086865-IW3_...
      reason: ready: 15 CSLCs + 1 CCSLCs; awaiting CCSLCs at 20170729
```

Cycle 417, the next acquisition again, is identical -- `complete=True`, all 27 referenced, `save_cc=False`. Both post-boundary dates close the round-trip.

This is the assertion that matters most, because it is where a disagreement between the download side's epoch and the evaluator's would surface. Each post-boundary acquisition builds its window on all 27 of the boundary's outputs and correctly marks itself `save_cc=False`, not being a boundary itself.

**Simulation caveat, stated so it is not over-read.** The venue runs `PGE_SIMULATION_MODE`, so SCIFLO emits fixture output: the 27 CCSLCs carry T041 burst ids and the L3 granule id names a different frame and epoch, where frame 31241 has two T117 bursts. Burst identities and counts here are artifacts and are **not** offered as evidence. What is real, and what this change owns, is everything upstream of the PGE: cycle numbering, frame attribution, window arithmetic, and the cascade reaching the PGE at all.

### Scenario B — phased/gap regression, unchanged behaviour

`collinss-0005`, frame 17235, unmodified database, exercising historical -> forward -> 2.8-year gap -> historical -> forward.

```
L3 = 31/31        CCSLC = 54/54        (datasets_e2e.json expectations)
SUCCESS: phases historical_01[15], forward_01[1], historical_02[15], forward_02[2], no_run[9]
SUCCESS: compressed CSLC boundaries at ['20170906', '20210804']
SUCCESS: 9 no_run date(s) produced no products
SUCCESS: every historical/no_run phase KSC is superseded
```

### Plain-frame regression

`gmanipon-0037`, frame 31241 on the **unmodified** database, anchored at its own first acquisition — `0, 210, 222, 234 …` — against `111, 135, 159 …` for the same frame stripped. Dated frames are untouched; only sensing-datetime-less frames change. 39 L3, 5 CCSLC, 11 SCIFLO jobs, zero failures. Cluster since destroyed.

### Three defects found along the way, none in this branch

**1. The all-phases smoke never sets the region whitelist.** `trigger-SCIFLO_L3_DISP_S1` gates on `metadata.region_id` against a whitelist defaulting to `["0"]`; the test frames are region 4. Isolating it by counting rule matches with and without that one clause:

```
collinss-0005:  without region clause -> 3   (exactly the 3 missing forward dates)
                with    region clause -> 0
```

The base DISP-S1 smoke calls `disp_s1_set_whitelist.py` four times (disable, set 4, set 0 for a negative test, disable); the all-phases smoke calls it **zero** times. Region gating arrived with OPERA-2617, already in this branch's base, so all-phases forward assertions fail on `develop` for the same reason. Deserves its own ticket.

**2. `run_disp_s1_forward_serial.py` cannot see products on a simulation-mode cluster.** Line 58 hardcodes `L3_INDEX = "grq_v1.0_l3_disp_s1*"`, but simulation-mode PGEs publish `v0.1`. The driver waits out its full L3 timeout and reports `FAIL ... L3=0` for products that exist — on exactly the configuration recommended for pre-PGE testing. The `fails` counts in this run are that false negative, not real failures.

**3. Re-ingesting a dataset across a monthly index rollover leaves a stale duplicate.** The same `_id` lands in the new month's index while the old copy remains, and wildcard queries see both. Here a stale `is_complete: false` copy in `-2026.08` outranked the repaired `is_complete: true` copy in `-2026.09` and silently held a k-window at 14/15.

**Also noted:** `tools/reevaluate_disp_s1_forward_kscs.py` refuses a sensing-datetime-less frame outright ("frame 31241 has no phases"). Consistent — no sensing datetimes means no phase labels — but it means this recovery tool cannot service the frames this ticket introduces.

### Not covered

Frame 26432 through the production `daac_data_subscriber.py query` entry point was not run. Real-PGE product content was not exercised, by design.

RKuttruff:
Merging now to start 6.0.6 release. We can run thorough testing for this during system testing & re-release as needed

PR #1491, opened 2026-08-28
Support querying of OPERA input products from GRQ

RKuttruff:
## Purpose
- Allows the querying of input products from a cluster's own GRQ records rather than relying on the DAACs/CMR

## Issues
- [OPERA-2553](https://hysds-core.atlassian.net/browse/OPERA-2553)

## Testing
- [x] Cluster Deploy
- [x] Testing
  - [x] DSWx-S1
    - [x] Manual
    - [x] 2026 Nepal flooding pre-/post-event S1D acquisition
  - [x] DISP-S1
    - [x] Data subscriber trigger
    - [x] PCM batch
    - [x] FWD (not needed as it's already driven by CSLCs in GRQ)
  - [x] DIST-S1
    - [x] HIST
    - [x] GRQ as Cache
- [ ] Regression checks/reviews
  - Before merging, I'd like for the following tagged devs to review the changes to the triggering workflows they worked on to sanity check that no regressions were introduced in CMR querying. Tests for GRQ querying will be handled at the system test level
  - [x] DSWx-S1 @chrisjrd 
  - [ ] DISP-S1 @pymonger 
  - [x] DIST-S1 @chrisjrd


data_subscriber/rtc_for_dist/dist_dependency.py
```diff
@@ -12,6 +12,20 @@
 
 GRQ_ES_DIST_S1_INDEX = "grq_*_l3_dist_s1*"
 CMR_RTC_CACHE_INDEX = "cmr_rtc_cache" #TODO: We should use wildcard later after we add year and month to the index name
+GRQ_INDEX_PATTERN = 'grq_*_l2_rtc_s1-*'
```
chrisjrd: consider renaming

data_subscriber/rtc_for_dist/dist_dependency.py
```diff
@@ -26,6 +40,16 @@ def file_paths_from_prev_product(previous_tile_product):
 
     return previous_tile_product["_source"]["metadata"]["product_s3_paths"]
 
+
+def get_cache_index_and_prefix(settings):
+    if settings.get('DIST_S1', {}).get('USE_RTC_CACHE', False):
```
chrisjrd, line 45: it may be clearer in the implementation to name the setting something like BASELINE_SOURCE: "GRQ" (or "CMR_RTC_CACHE"), rather than use a boolean use_cache.

data_subscriber/rtc_for_dist/dist_dependency.py
```diff
@@ -119,7 +145,7 @@ def get_previous_tile_product(self, download_batch_id, acquisition_ts):
 
         should_query = []
         for burst_id in all_burst_ids:
-            should_query.append({"match": {"burst_id.keyword": burst_id}})
+            should_query.append({"match": {f"{cache_prefix}burst_id.keyword": burst_id}})
```
chrisjrd, line 148: space might be missing afer the expression. or is this intentional?

data_subscriber/rtc_for_dist/dist_dependency.py
```diff
@@ -200,21 +226,30 @@ def sanity_check_cmr_rtc_cache(self):
         """
         Perform sanity check on the cmr_rtc_cache index.
         """
+        cache_index, cache_prefix = get_cache_index_and_prefix(self.settings)
+
+        if cache_index == GRQ_INDEX_PATTERN:
+            msg = (' You may want to set DIST_S1.USE_RTC_CACHE to true and populate it with a survey before switching'
```
chrisjrd, line 232: this needs to be documented somewhere to prevent ever seeing this log message.

data_subscriber/rtc_for_dist/rtc_for_dist_query.py
```diff
@@ -143,8 +147,87 @@ def unique_latest_granules(self, granules):
...
+    def run_query(self):
+        # Copy from base to simplify switching this class's override of the cmr query to use the class's query func
 
-    def query_cmr(self, timerange, now: datetime):
+        query_dt = datetime.now()
+        now = datetime.now(timezone.utc).replace(tzinfo=None)
+        query_timerange: DateTimeRange = get_query_timerange(self.args, now)
+
+        self.query_func = self._get_query_func(use_async=True)
+        granules = self._run_query_func(query_timerange, now)
```
chrisjrd, line 158: To avoid the copy-pasta, we could have probably added *args **kwargs to the signature in the parent class and update to check `if kwargs["_run_query_func"]` or something was present.

data_subscriber/rtc_for_dist/rtc_for_dist_query.py
```diff
@@ -153,10 +236,13 @@ def query_cmr(self, timerange, now: datetime):
                 self.logger.info(f"{self.args.tile_filter=}")
                 rtc_native_id_patterns = rtc_native_id_patterns_from_tiles(self.args.tile_filter)
                 self.args.native_id_patterns = rtc_native_id_patterns  # NOTE: informal arg being added here
-                granules = asyncio.run(async_query_cmr(self.args, self.token, self.cmr, self.settings, timerange, now))
+                granules = asyncio.run(self.query_func(timerange, now))   # TODO: Verify modification to self.args
```
chrisjrd: this should be resolved before merging the PR

data_subscriber/cmr.py
```diff
@@ -53,6 +53,7 @@ class Provider(StrEnum):
     ASF_CSLC = "ASF-CSLC"
     ASF_CSLC_STATIC = "ASF-CSLC-STATIC"
     ASF_NISAR_GCOV = "ASF-NISAR-GCOV"
+    GRQ = "GRQ"
```
chrisjrd, line 56: this is too generic. maybe it works, but the convention here would imply that this should be name something like `GRQ_RTC` or `GRQ_DIST` or something.

data_subscriber/download.py
```diff
@@ -125,8 +126,11 @@ def extract_one_to_one(self, product: Path, settings_cfg: dict, working_dir: Pat
         return PurePath(dataset_dir)
 
     def download_product_using_s3(self, url, token, target_dirpath: Path, args) -> Path:
+        # This feels a bit hacky, but we need to know if we're downloading from an archive bucket or an
+        #  OPERA bucket, since for the latter, we cannot use temp DAAC creds
+        is_opera_local_bucket = urlparse(url).netloc.startswith("opera-")
 
-        if self.cfg["USE_DAAC_S3_CREDENTIALS"] is True:
+        if self.cfg["USE_DAAC_S3_CREDENTIALS"] is True and not is_opera_local_bucket:
```
chrisjrd: the opera check should come first, for better readability. based on the comment preceding this line.

data_subscriber/es_conn_util.py
```diff
@@ -98,6 +101,9 @@ def get_document_timestamp_min_max(es_conn, index, timestamp_field):
         }
     }
     response = es_conn.es.search(index=index, body=query)
-    latest_timestamp = response["hits"]["hits"][0]["_source"][timestamp_field]
+    latest_source = response["hits"]["hits"][0]["_source"]
+    for field in timestamp_field.split('.'):
+        latest_source = latest_source[field]
+    latest_timestamp = latest_source
```
chrisjrd: latest_source seems redundant. consider renaming it to `latest_timestamp`.

data_subscriber/grq_query.py
```diff
@@ -0,0 +1,297 @@
...
+
+GLOBAL_BBOX = "-180,-90,180,90"
+SUPPORTED_PRODUCT_TYPES = {
+    ProductType.RTC,
+    ProductType.CSLC
+}
+
+
+async def async_query_grq(args, index_pattern, settings, timerange: DateTimeRange, now: datetime, verbose=True) -> list:
+    if index_pattern is None:
```
chrisjrd, line 32: the check seems redundant, given that it is hard-coded elsewhere. (overly defense programming anti-pattern). Consider removing the check after your testing.

data_subscriber/grq_query.py
```diff
@@ -0,0 +1,297 @@
...
+        _grq_doc_to_granule(doc, args.collection)
+        for doc in scan(es_conn, query, index=index_pattern)
+        if not _is_cache_doc(doc)
+    ]
+
+    for granule in granules:
+        granule["filtered_urls"] = _filter_granules(granule, args)
+
+    logger.info(f'Query complete. Found {len(granules):,} granule(s)')
+    logger.info(json.dumps(granules[0] if granules else [], indent=2),)  # TODO: switch to debug
```
chrisjrd: don't forget to resolve this before merging, if still applicable.

chrisjrd: you'll also want to do a check for the log level to avoid evaluating the dumps call every time.

data_subscriber/grq_query.py
```diff
@@ -0,0 +1,297 @@
...
+    }
+
+    if collection == Collection.RTC_S1_V1:
+        granule_doc['polarization'] = doc['metadata']['polarizations']
+
+    return granule_doc
+
+
+def _select_urls_list(local_urls: list, archive_urls: list) -> list:
+    if archive_urls is None:
```
chrisjrd: more pythonic to use `if not archive_urls`, unless you really need the `None`. e.g. how would `[]` be treated?

data_subscriber/grq_query.py
```diff
@@ -0,0 +1,297 @@
...
+        must.append({
+            "range": {
+                "creation_timestamp": {
+                    "gte": _datetime_to_es_query_timestamp(start_date),
+                    "lte": _datetime_to_es_query_timestamp(end_date)
+                }
+            }
+        })
+
+    if args.native_id is not None and not (hasattr(args, 'native_id_patterns') and args.native_id_patterns):
```
chrisjrd, line 188: at a first glance, it seems that these 2 if blocks should be swapped (check for product type first over native-id). Maybe what is here is okay for now.

data_subscriber/grq_query.py
```diff
@@ -0,0 +1,297 @@
...
+            # extract burst ID from the native-ID, and find the 1-2 relevant MGRS burst sets containing that burst ID.
+            match_native_id = re.match(rtc_granule_regex, args.native_id)
+            burst_id = mbc_client.product_burst_id_to_mapping_burst_id(match_native_id.group("burst_id"))
+            native_ids = mbc_client.get_reduced_rtc_native_id_patterns(mgrs[mgrs["bursts"].str.contains(burst_id)])
+
+            if not native_ids:
+                raise ValueError(f"The supplied {args.native_id=} is not associated with "
+                                 f"any land-based MGRS tile collection.")
+
+            for nid in native_ids:
```
chrisjrd: it may take a second to infer that nid means native_id. Consider writing it out for clarity. You could alternatively use a name like `id_` which works for any `*_id` iterable name.

data_subscriber/grq_query.py
```diff
@@ -0,0 +1,297 @@
...
+                should.append({
+                    "wildcard": {
+                        "id.keyword": nid
+                    }
+                })
+        elif isinstance(args.native_id, list) or '&native-id[]=' in args.native_id:
```
chrisjrd, line 206: if the native-id was the first param, it wouldn't have the `&`. this may cause problems with this check and the split done further down.

data_subscriber/grq_query.py
```diff
@@ -0,0 +1,297 @@
...
+                    "id.keyword": pattern
+                }
+            })
+
+    if args.bbox != GLOBAL_BBOX:
+        # TODO: Need to test
+        try:
+            min_lon, min_lat, max_lon, max_lat = (float(c) for c in args.bbox.split(','))
+
+            if max_lat <= min_lat or max_lon <= min_lon:
```
chrisjrd: consider consolidating this logic with the existing `_validate_bounds` function in the code. (i'd suggest creating a ticket to do it later, but we don't have a formal way to track this kind of tech debt)

data_subscriber/grq_query.py
```diff
@@ -0,0 +1,297 @@
...
+                                "shape": {
+                                    "type": "polygon",  # Bad GeoJSON, but it's what Tosca does...
+                                    "coordinates": [coords]
+                                }
+                            }
+                        }
+                    }
+                }
+            })
+        except Exception as e:
```
chrisjrd: this block can be reduced to just the portion that can throw the exception. I recommend reducing the scope of the block to make further refactor more obvious, and for readability reasons.

data_subscriber/grq_query.py
```diff
@@ -0,0 +1,297 @@
...
+            raise ValueError(f'Invalid bounding box: {args.bbox}') from e
+
+    if len(must) > 0:
```
chrisjrd, line 273: you can technically have `must: []` and `should: []` without adversely affecting the query. This is the rational for the `get_body()` implementation elsewhere in the code. This should help you simplify not only this block but the if-else blocks that assign to must/should.

tests/unit/test_dist_s1_input_tool.py
```diff
@@ -1173,6 +1173,7 @@ async def test_baseline_products_with_real_cmr_data(self):
 
         print("\n✓ Test passed: Baseline product selection works correctly with real CMR data")
 
+    @pytest.mark.xfail(reason='Needs updating')
```
chrisjrd, line 1176: i recommend creating a ticket for this, as it otherwise hides the effort to fully implement this change.

tools/populate_cmr_rtc_cache.py
```diff
@@ -15,14 +15,77 @@
...
+                "product_version": granule["product_version"],
+                "acquisition_cycle": granule["acquisition_cycle"],
+                "creation_timestamp": now
+            },
+            "dataset": "L2_RTC_S1_CACHED",
+            "ipath": "hysds::data/L2_RTC_S1_CACHED",
+            "system_version": "v1.0",
+            "dataset_level": "L2",
+            "dataset_type": "L2_RTC_S1_CACHED",
+            "version": "v1.0",
```
chrisjrd, line 82: the version fields may need to be sourced from somewhere to be foolproof. This may be fine for a while, but we should bring it up to Gerald or Hyun.

chrisjrd (approved):

chrisjrd:
consider flattening the commit history, as some commit changes add noise to the general commit history.

data_subscriber/rtc_for_dist/dist_dependency.py
RKuttruff: this is intentional

data_subscriber/rtc_for_dist/rtc_for_dist_query.py
RKuttruff: Done

data_subscriber/cmr.py
RKuttruff: That seems a bit excessive to me, cause then we'd have `GRQ_DSWX`, `GRQ_DIST` and `GRQ_DISP`

I think the reverse should be true: the providers list should _just_ be providers (`LPCLOUD`, `ASF`, `DATASPACE`, `GRQ`, `PODAAC`) and the product should be determined by other means, but I realize this is probably too much of a change to implement in this PR

data_subscriber/grq_query.py
RKuttruff: This could mainly be a development safeguard. For example if a new query type is introduced and the dev forgets to add an index pattern, they'd get a clear message

data_subscriber/grq_query.py
RKuttruff: removed the line entirely as it's no longer needed

data_subscriber/grq_query.py
RKuttruff: The functions that build the `native-id[]` params in this case all have the param be not-first. 

Eg: build_rtc_native_ids in dist_s1_utils line 149

data_subscriber/grq_query.py
RKuttruff: moved the validation to parser.py and thus am removing the entire try-except

tests/unit/test_dist_s1_input_tool.py
RKuttruff: There is a ticket for fixing the unit tests. I put this as xfail because it's out of date, so we know it needs updating in the future

tools/populate_cmr_rtc_cache.py
RKuttruff: These are not intended to be "proper" granules, but rather a temporary bootstrap for the GRQ data for DIST triggering until some proper RTC-fwd data can be populated, then it can(should) be simply purged. Realistically, this shouldn't even be needed in OPS since we're preserving the GRQ indices, and thus this wouldn't be needed anymore. I've also got an updated version of this in #1495 which should be favored.

@hhlee445 or @pymonger, please confirm this is ok?

RKuttruff:
Thanks, @chrisjrd I addressed a good handful of the review comments. Some others I'm not sure about and think need more discussion and/or require work out of this PR's scope (especially if we want to start the release today)

RKuttruff:
as for the commits, I tried squashing some but got an error and didn't feel comfortable pushing further

PR #1503, opened 2026-09-17
Use EDL bearer token for DIST-S1 confirmation tool COG access

riverma:
## Summary

- Adds `--edl-token` CLI flag (with `EDL_TOKEN` env var fallback) to `dist_s1_confirmation.py`
- When provided, passes the bearer token to GDAL via `GDAL_HTTP_HEADERS`, bypassing the implicit `~/.netrc` credential flow for COG file access
- Falls back to existing `~/.netrc` behavior when no token is given (backward-compatible)
- Updates README to document token-based auth as the preferred approach

## Motivation

During security review of nasa/opera-sds-ops#68 (opera-launchpad tool registration), storing plaintext EDL username/password in `~/.netrc` was flagged as a security concern. EDL bearer tokens are the preferred credential type — they're time-limited, revocable, and don't require storing passwords on disk.

See: https://github.com/nasa/opera-sds-ops/pull/68#pullrequestreview-5239089076

## Test plan

- [x] `--help` shows the new `--edl-token` flag
- [x] CMR survey runs without regression (unauthenticated path unchanged)
- [x] Run with a real EDL token against PROD — CMR survey completes cleanly with bearer auth
- [x] Verify fallback still works when `--edl-token` is not provided (no token, no `.netrc` — CMR survey runs without error)

Note: PROD DIST-S1 collection currently has 0 granules, so end-to-end COG file access with bearer auth couldn't be exercised. 

cc @RKuttruff


RKuttruff (approved):
Looks good to me

There's one change you may want to implement since we're (temporarily) including an option to source an EDL token using the `.netrc` for the CMR queries that you may want to supersede if `--edl-token` is provided. That, said, this particular use of the EDL token will be eventually removed once the DIST collection is made public

PR #1505, opened 2026-09-22
OPERA-2646: Integrate HySDS v6.4.5 and enable job locking

pymonger:
## Summary

Integrates HySDS framework **v6.4.5** into opera-sds-pcm (from v6.4.3, taking the v6.4.4 hop OPERA skipped) and turns on HySDS **job locking** for every OPERA job worker. Over v6.4.3 the release delivers the job_failed duplicate/race fixes HC-640, HC-648, HC-651 and HC-652, plus sdscli 2.1.3's `numpy<2.0` / `contourpy<1.4.0` venv pins (root cause HC-653 is still open upstream).

Resolves [OPERA-2646](https://hysds-core.atlassian.net/browse/OPERA-2646)

## Changes

Two commits; the second can be dropped on its own.

### Integrate HySDS v6.4.5
- **`cluster_provisioning/*`**: `hysds_release` default `v6.4.3` -> `v6.4.5` in all seven places (dev-e2e template.tfvars + variables.tf, dev, dev-int variables.tf + override.tf, int_pst_ops_devcommon_override.tf, ops).
- **`setup.py`**: hysds `v3.3.2` -> `v3.3.4`. hysds_commons (2.4.1), osaka, chimera and prov_es are unchanged in this release, so opera-sds-lambdas needs no change.
- **`conf/sds/files/supervisord.conf.mozart`**: add the `[program:reap_orphaned_job_failed]` block from hysds v3.3.4. OPERA overrides this file, so without the block the orphan reaper (HC-648, HC-651) never runs on an OPERA mozart. It ships report-only, as upstream does; deleting stays opt-in via `--delete-orphans`.
- **`conf/sds/files/job_status.template`**: add `job.retry_count` and `job.job_info.retry_delete.{index,seq_no,primary_term}` (sdscli 2.1.2, HC-648). OPERA overrides this template too; it now matches sdscli 2.1.3 field for field.
- **`cluster_provisioning/modules/common/mozart.tf`**: refresh the lightweight-jobs comment (v6.4.5 ships v2.1.3).

### Enable HySDS job locking
- **`cluster_provisioning/modules/common/mozart.tf`**: write `ENABLE_JOB_LOCKING: true` to `~/.sds/config`. mozart and factotum render celeryconfig from the upstream hysds template, which already reads it.
- **`conf/sds/files/celeryconfig.py.tmpl.asg`, `.private_verdi`**: add the job-lock settings block from the hysds v3.3.4 template. Without it the autoscaling verdi workers fall back to the hysds default of `False` whatever the sds config says. **Only the lock block is added**, not a full rebase on the upstream template: the rest of it would change cache eviction, cache mount mode and container home paths on those workers, because its `CACHE_READ_ONLY` / `EVICT_CACHE` defaults (false) differ from the hysds code defaults (true) and OPERA does not set them.
- **`JOB_LOCK_MAX_EXTENSIONS` 4320 -> 5760** (36 h -> 48 h) in both overrides. Seven OPERA job specs allow a 133200 s (37 h) `time_limit`; past the ceiling the heartbeat stops and the lock lapses while the job is still running.
- **`tests/unit/test_job_lock_config.py`** (new): both overrides render the flag from the sds config as a Python bool, the lock budget outlasts every job spec's `time_limit`, and provisioning enables it.

### How job locking works

With `ENABLE_JOB_LOCKING` on, `hysds.job_worker.run_job` does this for every job (hysds 3.3.4, `hysds/lock.py`):

1. **Finished-redelivery check (HC-652).** If the broker redelivered the task and redis already records `job-completed` or `job-deduped` for its uuid, return without running and without writing a status.
2. **Stale-lock check.** Look up the lock for the job's payload id. That id is the job_status `_id`, so redeliveries and retries of one job share it. A lock not renewed for 3 heartbeats (90 s), or with no metadata, is broken. A recently renewed lock is watched for up to ~100 s to see whether its holder is still alive.
3. **Acquire** a Redlock `job-lock-<payload_id>` in mozart redis (wait up to 30 s), plus a metadata key recording task id, worker host and last renewal. If another execution holds it:
   - a redelivered copy records `job-deduped` and returns, leaving the live original running;
   - a non-redelivered execution fails with `Job <payload_id> already running (task ... on worker ...)`, or logs a `job_lock_contention` worker anomaly when a newer attempt already owns the job doc.
4. **Heartbeat.** A thread in the worker process extends the lock and its metadata every 30 s. The lock lapses 600 s after the last extension, and the extension budget is `JOB_LOCK_MAX_EXTENSIONS` (5760 here, 48 h).
5. **Release** in a `finally` when the job ends. A hard-killed worker cannot release; its lock goes stale after 90 s and the next delivery breaks it.

Without locking, a redelivered task whose job reads `job-started` sleeps 60 s, then re-runs unless the previous worker's redis worker-status key still exists. mozart's `process_events` refreshes that key with a 60 s TTL from each celery worker-heartbeat event, so it lapses whenever a worker's events stop reaching mozart over the broker, even while its job keeps running, and the job then runs twice. The lock's heartbeat is written to redis directly by the worker process that runs the job, so it depends on that process being alive, not on the broker.

It is not payload-hash dedup: distinct submissions have distinct payload ids, and `enable_dedup=false` reprocessing is unaffected. NISAR has run it since February and SWOT runs it in production. v6.4.5 is the first safe release: HC-652 was a locking-only defect where a redelivered, already finished job re-ran and lost its success record.

### What it changes on OPS FWD (measured read-only, 2026-09-23)

Over the job_status retention window (2026-09-07 to 2026-09-23, 6.7 M job docs), 4,762 SCIFLO executions ran on a broker redelivery. Matching each one to its products by producing job id:

| Redelivered execution | DSWx-HLS | DSWx-S1 | RTC-S1 | CSLC-S1 |
|---|---|---|---|---|
| No duplicate product | 3,587 | 793 | 321 | 25 |
| Duplicate product; original published, then died before recording completion | 33 | 0 | 1 | 0 |
| Duplicate product; original still running when the copy started | 2 | 0 | 0 | 0 |

- **Prevented by locking: the 2 concurrent runs.** Example: job `1c9b745e` on 2026-09-11. The redelivered copy started at 08:56:13Z, the still-running original published at 09:00:41Z and the copy published at 09:01:13Z, giving two DSWx-HLS products for one input granule. With the lock, the copy would have seen the original's heartbeat and deduped.
- **Not prevented: the 34 dead-original re-runs.** The original published and then died before writing `job-completed`, so its heartbeat stopped with it and the copy correctly took over. Because DSWx-HLS and RTC-S1 product ids carry the production time, the second publish is a new product rather than a no-clobber collision. This is a separate issue and not addressed here.

It does not address the other OPS FWD job-accounting issues:
- the concurrent-retry race (HC-629: two user rules fire two `lw-mozart-retry` runs, which race in `retry.py` on different payloads);
- job_failed duplicates and orphans (HC-640, HC-648, HC-651), which the framework bump in the first commit fixes, with the reaper removing the redelivery pair;
- the 4 `Dataset already exists` download failures in the window, which were separate download jobs (different payloads) fetching the same granule.

Nor does it reduce redeliveries themselves; it changes how a redelivered copy decides whether to run.

### Costs and behavior changes
- One heartbeat thread and one redis write per running job every 30 s.
- A redelivered copy of a hard-killed job waits up to ~100 s, instead of 60 s, before re-running.
- New failure signature for a non-redelivered contender: `Job <payload_id> already running`. It matches none of OPERA's retry rules, so it cannot start a retry loop.
- While a live original runs, a deduped copy writes `job-deduped` under the shared job id; the original's own terminal write replaces it.

## Documentation

- Release notes: [hysds-framework v6.4.5](https://github.com/hysds/hysds-framework/releases/tag/v6.4.5), [v6.4.4](https://github.com/hysds/hysds-framework/releases/tag/v6.4.4), [hysds v3.3.4](https://github.com/hysds/hysds/releases/tag/v3.3.4), [hysds v3.3.3](https://github.com/hysds/hysds/releases/tag/v3.3.3)
- Upstream: HC-640, HC-648, HC-651, HC-652 (delivered); HC-653 (open); job locking HC-590 / HC-609; retry race HC-629 (open)

## Test plan

- [x] `tests/unit/test_job_lock_config.py` + `test_job_spec_queues.py`: 77 passed. Negative control (upstream 4320 ceiling, flag removed) fails 3 as expected.
- [x] Both celeryconfig overrides render with Jinja and compile as Python; `ENABLE_JOB_LOCKING = True`, `JOB_LOCK_MAX_EXTENSIONS = 5760`.
- [x] All five `setup.py` component tarball URLs resolve; no `v6.4.3` / `v3.3.2` references remain.
- [x] v6.4.5 amd64 bundles mirrored to the OPERA S3 artifact mirror (5/5, size-verified)
- [x] `Verdi=v6.4.5` EBS snapshot baked (`snap-04342560b5a38d8c9`, Jenkins `opera-pcm-docker-ebs-snapshot-creator` #61)
- [x] `dev-e2e-pge-DISP_S1` on gmanipon-0043 (boundary-serial forward, like the develop baseline gmanipon-0041): apply rc=0 in 20h57m (baseline 21h37m); `check_pcm` 4/5, identical to the baseline
  - historical F31241 14 L3 + 2 CCSLC; phased historical F24718 14 L3 + 3 CCSLC with all five phase-structure checks SUCCESS
  - forward: 11 fired, 45 no-fire (1 superseded), 0 failed, the same as the baseline; frame 31241 ends at 26 L3
  - PGE-fed CSLC phase: 5/5 SUCCESS (burst metadata, CSC 2/2, KSC, L3 published, full lineage)
  - the one failure is `test_forward_expected_datasets`, whose `datasets_e2e.json` still expects 55 forward L3; it fails the same way on develop
  - job accounting: 451 completed, 0 failed, no `job_failed` index; 3 redelivered executions each completed once; 0 lock-contention events; 0 locks left in redis at the end
- [ ] `dev-e2e-pge-DISP_S1-all_phases` on mplotkin-0005
- [ ] Post-deploy: hysds 3.3.4 / hysds_commons 2.4.1 / numpy<2 in the venvs; lightweight-jobs v2.1.3 imported; `reap_orphaned_job_failed` RUNNING; `ENABLE_JOB_LOCKING = True` rendered for autoscaling verdi and factotum, and locks acquired in job logs; no duplicate or orphaned job_failed documents


hhlee445 (approved):
Ran regression tests and all PGEs executed successfully.

PR #1507, opened 2026-09-22
Attempt to derive S3 URL before using HTTP

RKuttruff:
Should avoid issues with trying to determine PODAAC S3 URLs by parsing the CMR JSON as provided URL for the CMR JSON is inaccessible over HTTP from the verdi VPC, but should be of the form `https://<hostname>/<s3 bucket name>/<s3 key>`, meaning its corresponding S3 URL can be derived.

## Issues
- [OPERA-2648](https://hysds-core.atlassian.net/browse/OPERA-2648)


hhlee445 (approved):
LGTM

